PR #1, opened 2021-02-15
Support cyclic mover navigation

ling1726:
* Changes mover API to object instead of enum
* Change the typing to use enum instead of interface+const
* Add cyclic navigation check to `FocusedElement` API

### DIFF
--- a/core/src/State/FocusedElement.ts
+++ b/core/src/State/FocusedElement.ts
@@ -417,9 +417,9 @@ export class FocusedElementState
 
                 // cyclic navigation, focus first or last elements in the mover container respectively
                 if (isPrev) {
-                    next = ctx.mover.lastElementChild as HTMLElement;
+                    next = this._ah.focusable.findLast(ctx.mover);
                 } else {
-                    next = ctx.mover.firstElementChild as HTMLElement;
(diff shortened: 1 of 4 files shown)


core/src/State/FocusedElement.ts
```diff
@@ -410,7 +410,17 @@ export class FocusedElementState
                 // Nowhere to move inside the current Mover.
                 e.preventDefault(); // We don't need the page to scroll when we're custom-handling
                                     // the arrows.
-                return;
+
+                if (!ctx.moverOptions?.cyclic) {
+                    return;
+                }
+
+                // cyclic navigation, focus first or last elements in the mover container respectively
```
mshoho, line 418: This is assuming that first/last focusable are first/last direct descendants of the container. Which mostly won't be the case in the real scenarios.

There is an example how to find the first/last focusable inside the container above:
```? (isPrev // Find next focusable outside of the Mover container.
                    ? this._ah.focusable.findFirst(ctx.mover)
                    : this._ah.focusable.findLast(ctx.mover)```

ling1726: Ah i didn't know there was an API in focusable for this, will modify

PR #4, opened 2021-02-16
Encourage single AH instance on window

ling1726:
* Adds internal window instance of AH on creation
* Added  `abilityHelpersExists` helper method
* Throw error on `createAbilityHelpers` if an AH instance already exists on the window

### DIFF
--- a/core/src/AbilityHelpers.ts
+++ b/core/src/AbilityHelpers.ts
@@ -177,7 +177,7 @@ export function createAbilityHelpers(win: Window, props?: Types.AbilityHelpersCo
     const ah = new AbilityHelpers(win, props);
     // @ts-ignore
     win.BE_ABLE_INSTANCE = ah;
-    return ah
+    return ah;
 }
 
 export function getOutline(ah: Types.AbilityHelpersCore): Types.OutlineAPI {
@@ -279,4 +279,13 @@ export function abilityHelpersExists(win: Window): boolean {
 
     // @ts-ignore
     return !!win.BE_ABLE_INSTANCE;
+}
+
+export function getWindowInstance(win: Window): Types.AbilityHelpersCore | null {
+    if (abilityHelpersExists(win)) {
+        // @ts-ignore
+        return win.BE_ABLE_INSTANCE;
+    }
+
+    return null;
 }
--- a/core/src/AbilityHelpers.ts
+++ b/core/src/AbilityHelpers.ts
@@ -182,15 +182,17 @@ class AbilityHelpers implements Types.AbilityHelpersCore, Types.AbilityHelpersIn
 /**
  * Creates an instance of ability helpers, returns the current window instance if it already exists
  */
-export function createAbilityHelpers(win: WindowWithAHInstance, props?: Types.AbilityHelpersCoreProps): Types.AbilityHelpersCore {
-    const existingAh = getCurrentAbilityHelpers(win);
+export function createAbilityHelpers(win: Window, props?: Types.AbilityHelpersCoreProps): Types.AbilityHelpersCore {
+    const existingAh = getCurrentAbilityHelpers(win as WindowWithAHInstance);
     if (existingAh) {
-        console.warn('Attempted to create a duplicate ability helpers instance on the window');
+        if (__DEV__) {
+            console.warn('Attempted to create a duplicate ability helpers instance on the window');
+        }
         return existingAh;
     }
 
     const ah = new AbilityHelpers(win, props);
-    win.__ahInstance= ah;
+    (win as WindowWithAHInstance).__ahInstance = ah;
     return ah;
 }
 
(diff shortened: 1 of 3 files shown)


core/src/AbilityHelpers.ts
```diff
@@ -170,7 +170,14 @@ class AbilityHelpers implements Types.AbilityHelpersCore, Types.AbilityHelpersIn
 }
 
 export function createAbilityHelpers(win: Window, props?: Types.AbilityHelpersCoreProps): Types.AbilityHelpersCore {
-    return new AbilityHelpers(win, props);
+    if (abilityHelpersExists(win)) {
+        throw new Error(' An Ability helpers instance already exists on this window, you might use `abilityHelpersExists` to validate creation');
+    }
+
+    const ah = new AbilityHelpers(win, props);
+    // @ts-ignore
```
mshoho, line 178: Let's not use ts-ignore. Also all internal things ability-helpers stores on the elements are prefixed __ah.
Look at the WindowWithUID example in Utils.ts.
Let's define WindowWithAbilityHelpersInstance with __ahInstance prop. Makes sense to not make it exported, but to make internally `(win as WindowWithAbilityHelpersInstance).__ahInstance = ah;`. That way it's less error-prone to the future refactorings.

core/src/AbilityHelpers.ts
```diff
@@ -264,3 +271,21 @@ export function getAbilityHelpersAttribute(
         [Types.AbilityHelpersAttributeName]: attr
     };
 }
+export function abilityHelpersExists(win: Window): boolean {
```
mshoho, line 274: Probably instead of two methods (abilityHelpersExists and getWindowInstance) makes sense to do just getCurrentAbilityHelpers(win).

core/src/AbilityHelpers.ts
```diff
@@ -170,7 +170,14 @@ class AbilityHelpers implements Types.AbilityHelpersCore, Types.AbilityHelpersIn
 }
 
 export function createAbilityHelpers(win: Window, props?: Types.AbilityHelpersCoreProps): Types.AbilityHelpersCore {
```
mshoho, line 172: Also we need to delete the instance when disposeAbilityHelpers is called.

ling1726: Added this to `dispose` member of `AbilityHelpers`

core/src/AbilityHelpers.ts
ling1726: Yeah that's a good call, done

ling1726: Also log a warning and return an existing instance if it exists, error could be a bit too far

core/src/AbilityHelpers.ts
ling1726: Added an internal (non-exported) interface

core/src/AbilityHelpers.ts
```diff
@@ -169,8 +179,16 @@ class AbilityHelpers implements Types.AbilityHelpersCore, Types.AbilityHelpersIn
     }
 }
 
-export function createAbilityHelpers(win: Window, props?: Types.AbilityHelpersCoreProps): Types.AbilityHelpersCore {
-    return new AbilityHelpers(win, props);
+export function createAbilityHelpers(win: WindowWithAHInstance, props?: Types.AbilityHelpersCoreProps): Types.AbilityHelpersCore {
```
mshoho, line 185: That's the public API, it should just have Window in the signature, let's do `win as WindowWithAHInstance` internally in the function.

core/src/AbilityHelpers.ts
```diff
@@ -264,3 +282,12 @@ export function getAbilityHelpersAttribute(
         [Types.AbilityHelpersAttributeName]: attr
     };
 }
+
+
+export function getCurrentAbilityHelpers(win: WindowWithAHInstance): Types.AbilityHelpersCore | null {
+    if (win && win.__ahInstance) {
```
mshoho, line 295: You use optional chaining above, why not use it here like `return win?.__ahInstance || null` for more compactness.

core/src/index.ts
```diff
@@ -12,5 +12,6 @@ export {
     getModalizer,
     getObservedElement,
     getOutline,
+    abilityHelpersExists,
```
mshoho, line 15: This function doesn't seem to exist anymore, does the project build like that?

core/src/AbilityHelpers.ts
```diff
@@ -169,8 +179,19 @@ class AbilityHelpers implements Types.AbilityHelpersCore, Types.AbilityHelpersIn
...
 }
 
-export function createAbilityHelpers(win: Window, props?: Types.AbilityHelpersCoreProps): Types.AbilityHelpersCore {
-    return new AbilityHelpers(win, props);
+/**
+ * Creates an instance of ability helpers, returns the current window instance if it already exists
+ */
+export function createAbilityHelpers(win: WindowWithAHInstance, props?: Types.AbilityHelpersCoreProps): Types.AbilityHelpersCore {
+    const existingAh = getCurrentAbilityHelpers(win);
+    if (existingAh) {
```
mshoho, line 187: Makes sense to add `if (__DEV__)` to the console log. So that it's not affection the production bundle.

core/src/AbilityHelpers.ts
ling1726: it's a reflex  when writing if conditions, will remove

core/src/AbilityHelpers.ts
ling1726: yeah you're right

PR #7, opened 2021-03-01
Fix modalizer examples  add additional logging

ling1726:
This PR fixes the modalizer examples by initializing the Modalizer and
Deloser APIs before the stories are rendered. This was not caught
previously during the migration from `demo.tsx` beceause those APIs were
initialized on the global scope.

Also adds some additional documentation and console.error development
output to warn about this in the future

### DIFF
--- a/examples/src/Modalizer.stories.tsx
+++ b/examples/src/Modalizer.stories.tsx
@@ -1,25 +1,11 @@
 import * as React from 'react';
 import { Meta } from '@storybook/react';
-import { getAbilityHelpersAttribute, getCurrentAbilityHelpers, getDeloser, getModalizer } from 'ability-helpers';
+import { getAbilityHelpersAttribute } from 'ability-helpers';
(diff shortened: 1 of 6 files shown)


examples/src/Modalizer.stories.tsx
```diff
@@ -1,11 +1,26 @@
...
+    title: 'Modalizer',
+    decorators: [
+        (Story) => {
+            const ah = getCurrentAbilityHelpers(window);
+            
+            // These APIs need to initialized before use, no reason to do it for all stories yet
+            if (ah) {
+                getModalizer(ah);
+                getDeloser(ah);
+            }
```
ling1726, line 17: @mshoho  Should all APIs be initialized in examples by default ?

mshoho: To reduce the hustle, they could certainly be initialized by default. The reason for them to be initalized separately is the treeshekability.

PR #8, opened 2021-03-02
Add popup content example with modalizer

ling1726:
Adds an example that leverages the modalizer to render a button with a popup focus trap:

ucZ32kvzl6

### DIFF
--- a/core/src/Instance.ts
+++ b/core/src/Instance.ts
@@ -203,6 +203,7 @@ export function updateAbilityHelpersByAttribute(
         }
     }
 
+    debugger;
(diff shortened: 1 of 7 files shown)


core/src/Instance.ts
```diff
@@ -203,6 +203,7 @@ export function updateAbilityHelpersByAttribute(
         }
     }
 
+    debugger;
```
mshoho, line 206: Oops.

ling1726: Oops... removed

```suggestion
```

PR #10, opened 2021-03-30
Handle shift+tab nested movers

ling1726:
We consider a mover to be a single tab stop, therefore when focus is inside the mover:
1. TAB - should move to the next focusable element outside the mover
2. SHIFT+TAB - should mover the previous focusable element outside the mover

`1.` is already handled by default since walking the tree will always lead to the boundary of the most nested mover

In order to handle `2.` the `FocusedElementState` needs to traverse up the tree to until there is no more mover to determine what is the root mover element.

### DIFF
--- a/core/src/State/FocusedElement.ts
+++ b/core/src/State/FocusedElement.ts
@@ -383,7 +383,7 @@ export class FocusedElementState
                 if (isPrev) {
                     let  parentCtx: typeof ctx | undefined = ctx;
                     let rootMover = ctx.mover;
-                    while(parentCtx?.mover && parentCtx?.mover.parentElement) {
+                    while(parentCtx?.mover?.parentElement) {
                         rootMover = parentCtx.mover;
                         parentCtx = RootAPI.getTabsterContext(this._tabster, parentCtx.mover.parentElement);
                     }
--- a/core/src/State/FocusedElement.ts
+++ b/core/src/State/FocusedElement.ts
@@ -390,6 +390,7 @@ export class FocusedElementState
 
                     fromElement = this._tabster.focusable.findFirst(rootMover);
                 } else {
+                    // no need to find root since tree walking will always find the most nested mover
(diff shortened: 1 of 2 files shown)


core/src/State/FocusedElement.ts
```diff
@@ -375,12 +375,26 @@ export class FocusedElementState
...
-                : curElement;
+            let fromElement: HTMLElement | null = curElement;
+
+            // If the current element is in a mover, move to the mover boundaries since a mover is considered a single tabstop
+            if (isTab && ctx.mover && ctx.moverOptions?.navigationType === Types.MoverKeys.Arrows) {
+                // Consider nested movers a as a single tab stop, go up until there is no more mover
+                if (isPrev) {
+                    let  parentCtx: typeof ctx | undefined = ctx;
+                    let rootMover = ctx.mover;
+                    while(parentCtx?.mover && parentCtx?.mover.parentElement) {
```
mshoho, line 386: You already use optional chaining, `while (parentCtx?.mover?.parentElement)` should do it.

core/src/State/FocusedElement.ts
```diff
@@ -375,12 +375,26 @@ export class FocusedElementState
...
+                        rootMover = parentCtx.mover;
+                        parentCtx = RootAPI.getAbilityHelpersContext(this._ah, parentCtx.mover.parentElement);
+                    }
+
+                    fromElement = this._ah.focusable.findFirst(rootMover);
+                } else {
+                    fromElement = this._ah.focusable.findLast(ctx.mover);
```
mshoho, line 393: It seems to me that the same root finding logic should be applied for the moving forward :).

ling1726: It's not really necessary, since the way the tree walking works, it will always find the correct boundary element

ling1726: I've added a comment to reflect that

ling1726: Ah i see your point, donw

PR #12, opened 2021-03-30
Focusable can optioanlly ignore aria-disabled

ling1726:
According to [WAI](https://www.w3.org/TR/wai-aria-practices/#kbd_disabled_controls) spec some `aria-disabled` elements should still be
focusable

This PR adds an `ignoreAriaDisabled` option for focusable so that
Tabster's `isAccessible` check allways passes if set

### DIFF
--- a/core/src/Focusable.ts
+++ b/core/src/Focusable.ts
@@ -975,12 +975,14 @@ export class FocusableAPI implements Types.FocusableAPI {
             newProps.isDefault = props.isDefault;
             newProps.isIgnored = props.isIgnored;
             newProps.mover = props.mover;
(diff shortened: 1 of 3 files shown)


core/src/Focusable.ts
```diff
@@ -1044,8 +1046,6 @@ export class FocusableAPI implements Types.FocusableAPI {
                 return false;
             }
 
-            attrVal = e.getAttribute('aria-disabled');
```
mshoho, line 1047: Something is fishy here. I suppose you were willing to check the new prop, but have just removed the line unconditionally.

PR #17, opened 2021-04-08
chore: introduce GH actions

layershifter:
# WIP


package.json
```diff
@@ -5,6 +5,7 @@
     "build": "lerna exec -- npm run build",
     "clean": "lerna exec -- npm run clean && lerna clean --yes",
     "start": "lerna exec --scope examples -- npm run start",
+    "lint": "npm run tslint:check",
```
layershifter, line 8: This will include Prettier in next PRs.

PR #19, opened 2021-04-09
RTL awareness for tabster context and arrow keys

### LINKED ISSUE #5: Support RTL for Mover API
AH needs extra configuration to be RTL aware so that mover can know when to flip left/right arrow keys

AH needs to be able to configure RTL state at runtime

ling1726:
This PR adds RTL checking for `getTabsterContext` through an extra
option, that checks all ancestors in the dom tree for the `dir`
attribute.

Also uses this new RTL flag to flip arrow keys (left/right) for groupper
and mover use cases.

Resolves #5

### DIFF
--- a/core/src/Root.ts
+++ b/core/src/Root.ts
@@ -542,7 +542,18 @@ export class RootAPI implements Types.RootAPI {
         return _rootById[id];
     }
 
-    static getTabsterContext(tabster: Types.TabsterCore, element: Node): Types.TabsterContext | undefined {
+    /**
+     * Fetches the tabster context for an element walking up its ancestors
+     * 
+     * @param tabster Tabster instance
+     * @param element The element the tabster context should represent
+     * @param options Additional options
+     * @returns undefined if the element is not a child of a tabster root, otherwise all applicable tabster behaviours and configurations
+     */
+    static getTabsterContext(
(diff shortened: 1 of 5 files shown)


core/src/Root.ts
```diff
@@ -553,24 +564,26 @@ export class RootAPI implements Types.RootAPI {
...
             if (!tabsterOnElement) {
+                curElement = curElement.parentElement;
                 continue;
             }
 
             if (!groupper && tabsterOnElement.groupper) {
                 groupper = tabsterOnElement.groupper;
             }
 
-            const cfk = tabsterOnElement.focusable?.mover;
```
mshoho, line 568: There were two points to have this const: it minifies almost twice as better and (yeah, pretty little but still) is saves from the secondary lookup.

ling1726: alright, can we have a better name than `cfk` ? like `moverElement` ?

mshoho: I believe cfk was initially standing for currentFocusKey (or something like that). I suppose moverOnElement (or moe :)) would be consistent with tabsterOnElement :).

ling1726: moverOnElement it is :)

PR #25, opened 2021-04-28
FocusedElement: capture keydown listener

ling1726:
Runs Tabster's `keydown` listener for a focused element before the rest
of the window.

Ensures that `Tabster` behaviour runs before user behaviour
i.e.: Tabster drags you into a pit of success and it's up to you to
climb out

### DIFF
--- a/core/src/State/FocusedElement.ts
+++ b/core/src/State/FocusedElement.ts
@@ -97,10 +97,10 @@ export class FocusedElementState
             this._initTimer = undefined;
         }
 
-        win.document.removeEventListener('focusin', this._onFocusIn, true); // Capture!
-        win.document.removeEventListener('focusout', this._onFocusOut, true); // Capture!
-        win.document.removeEventListener('mousedown', this._onMouseDown, true); // Capture!
-        win.removeEventListener('keydown', this._onKeyDown);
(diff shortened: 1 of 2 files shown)


core/src/State/FocusedElement.ts
```diff
@@ -78,10 +78,11 @@ export class FocusedElementState
...
 
-        win.document.addEventListener('focusin', this._onFocusIn, true); // Capture!
-        win.document.addEventListener('focusout', this._onFocusOut, true); // Capture!
-        win.document.addEventListener('mousedown', this._onMouseDown, true); // Capture!
-        win.addEventListener('keydown', this._onKeyDown);
+        // Add these event listeners as capture - we want Tabster to run before user event handlers
+        win.document.addEventListener('focusin', this._onFocusIn, true);
+        win.document.addEventListener('focusout', this._onFocusOut, true);
+        win.document.addEventListener('mousedown', this._onMouseDown, true);
+        win.addEventListener('keydown', this._onKeyDown, true);
```
layershifter, line 85: Should it be also `true` on line 103?

ling1726: yes it should

PR #26, opened 2021-05-08
Unique modalizer instance per user id

ling1726:
This PR refactors modalizer so that one single instance exists and
manages and Modalizer DOM elements per user id. This makes sense since
all modalizers with the same id need to behave the same way.

This will enable more logic to be moved from `Root` to `Modalizer` in
the future. Lots of logic currently in `Root` regarding modalizer exists
because of the limitation that multiple modalizer instances can exist
that have the same user id.


ling1726:
@mshoho would welcome any feedback

PR #27, opened 2021-05-10
Remove modalizer accessibility and set aria-hidden in tree instead

ling1726:
Removes the concept of `accessible` for `Modalizer`

`Modalizer` will set `aria-hidden` on all non-Modalizer content on the document when active, and restore `aria-hidden` when no longer active.

Should result in most `Modalizer` references no longer needed for focusing/finding

### DIFF
--- a/core/src/Deloser.ts
+++ b/core/src/Deloser.ts
@@ -168,12 +168,12 @@ class DeloserHistoryByRoot extends DeloserHistoryByRootBase<Types.Deloser, Delos
         //     }
         // }
 
-        // // Nothing is found, at least try to reset.
-        // for (let id of Object.keys(resetQueue)) {
-        //     if (await resetQueue[id].resetFocus()) {
-        //         return true;
-        //     }
-        // }
+        // Nothing is found, at least try to reset.
+        for (let id of Object.keys(resetQueue)) {
+            if (await resetQueue[id].resetFocus()) {
+                return true;
+            }
+        }
 
         return false;
     }
--- a/core/src/Deloser.ts
+++ b/core/src/Deloser.ts
@@ -200,7 +200,7 @@ export class DeloserHistory {
 
         const historyByRoot = this.make(rootUId, () => new DeloserHistoryByRoot(this._tabster, rootUId));
 
-        if (!ctx || !ctx.modalizer || (ctx.root.getCurrentModalizerId() === ctx.modalizer.userId)) {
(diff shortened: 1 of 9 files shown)


core/src/Deloser.ts
```diff
@@ -196,7 +202,7 @@ export class DeloserHistory {
 
         const historyByRoot = this.make(rootUId, () => new DeloserHistoryByRoot(this._tabster, rootUId));
 
-        if (!ctx || !ctx.modalizer || (ctx.root.getCurrentModalizerId() === ctx.modalizer.userId)) {
+        if (!ctx || !ctx.modalizer?.isActive()) {
```
ling1726, line 205: this is not the same as previous condition

core/src/Focusable.ts
```diff
@@ -1273,20 +1284,16 @@ export class FocusableAPI implements Types.FocusableAPI {
         ignoreGroupper?: boolean
     ): number {
         const ctx = RootAPI.getTabsterContext(this._tabster, element);
-        const currentModalizerId = ctx && ctx.root.getCurrentModalizerId();
+        if (this._isHidden(element)) {
```
ling1726, line 1287: Should this be `isAccessible` too ? 
```suggestion
        if (this.isAccessible(element)) {
```

core/src/State/FocusedElement.ts
```diff
@@ -727,48 +710,8 @@ export class FocusedElementState
         return pde;
     }
 
-    private _validateFocusedElement = (element: HTMLElement, details: Types.FocusedElementDetails): void => {
```
ling1726, line 730: Let's check that this removal was done correctly

core/src/Modalizer.ts
```diff
@@ -381,40 +447,97 @@ export class ModalizerAPI implements Types.ModalizerAPI {
...
+        // Modalizer behaviour is opt in, only apply to elements that have a tabster context
+        if (!ctx) {
+            return;
         }
 
-        if (modalizer) {
-            if (this._curModalizer && (modalizer !== this._curModalizer)) {
-                this._curModalizer.setFocused(false);
-            }
+        if (this._focusOutTimer) {
```
ling1726, line 487: the `focusOutTimer` shouldn't be necessary anymore

core/src/State/FocusedElement.ts
ling1726: Previously this might have caused a bug for `isOthersAccesible`

PR #29, opened 2021-05-20
chore(keyborg): Cleaup and dedupe

ling1726:
Adds docstrings and renames some things for readability

Dedupes multipe `WeakRef` classes by creating a weak ref generic wrapper

### DIFF
--- a/keyborg/src/FocusEvent.ts
+++ b/keyborg/src/FocusEvent.ts
@@ -2,25 +2,32 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
+import { WeakRefInstance } from './WeakRefInstance';
 
 export const KEYBORG_FOCUSIN = 'keyborg:focusin';
 
-interface CustomFocusFunctionWithOriginal {
-    __keyborgFocus?: (options?: FocusOptions | undefined) => void;
+interface KeyborgFocus {
+    /**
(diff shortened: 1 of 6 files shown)


keyborg/src/WeakRefInstance.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        if (this._weakRef) {
+            instance = this._weakRef?.deref();
+
+            if (!instance) {
+                delete this._weakRef;
+            }
+        } else {
+            instance = this._instance;
+            if ((instance as Disposable)?.isDisposed?.()) {
```
mshoho, line 47: I'm wondering what is this dot after isDisposed?

keyborg/src/FocusEvent.ts
```diff
@@ -80,93 +56,108 @@ export interface KeyborgFocusInEvent extends Event {
...
 
-    wnd.HTMLElement.prototype.focus = focus;
+    kwin.HTMLElement.prototype.focus = focus;
 
-    const data: KeyborgFocusEventData = wnd.__keyborgData = {
+    const data: KeyborgFocusEventData = kwin.__keyborgData = {
         focusInHandler: (e: FocusEvent) => {
             const target = e.target as HTMLElement;
+            if (!target) {
+                return;
```
mshoho, line 95: By the way, from the minification point of view `if(condition){ actions }` is a few bytes shorter than `if(!condition){return} actions` :)

PR #31, opened 2021-05-24
Fix(autoRoot): auto root should have tabster attr

ling1726:
Currently `autoRoot` does not have tabster attributes applied to the DOM
element. This results in mutations not being dispatched to the `RootAPI`
because `getTabsterOnElement` is used on the mutation target before the
observer decides to dispatch the mutation event

### DIFF
--- a/core/src/Root.ts
+++ b/core/src/Root.ts
@@ -90,6 +90,11 @@ export class Root implements Types.Root {
             delete this._dummyInputLast;
         }
 
+        const rootElement = this._element.get();
+        if (rootElement) {
+            setTabsterOnElement(this._tabster, rootElement, { root: undefined });
+        }
+
         this._forgetFocusedGrouppers = () => {/**/};
     }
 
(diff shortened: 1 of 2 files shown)


core/src/Root.ts
```diff
@@ -479,6 +479,8 @@ export class RootAPI implements Types.RootAPI {
                         rootAPI._forgetFocusedGrouppers,
                         rootAPI._autoRoot
                     );
+
+                    setTabsterOnElement(rootAPI._tabster, body, { root: rootAPI._autoRootInstance });
```
mshoho, line 483: Makes sense to add a cleanup for that in the dispose method above. To not leave the data attribute residue in the DOM when everything's disposed.

ling1726: I've added this to `Root` instance's `dispose` method instead of `RootAPI`. WDYT ?

mshoho: I suppose we need to go deeper™. For the consistency let's do setTabsterOnElement in the constructor (that also means removing it from  add too).

And perhaps let's remove the move() method completely. It's never used at this point. And if we're going to have a use case for it, we'll add a better solution depending on that use case.

ling1726: Agreed, it's looking a lot better now, only the `Root` instance uses `setTabsterOnElement` 👍👍👍

PR #35, opened 2021-05-31
Accessible option for waiting observed element.

mshoho:


core/src/CrossOrigin.ts
```diff
@@ -687,7 +689,7 @@ class GetElementTransaction extends CrossOriginTransaction<CrossOriginElementDat
                     let isForwardResolved = false;
                     let isResolved = false;
 
-                    (tabster as unknown as Types.TabsterInternal).observedElement!!!.waitElement(name, timeout).then(value => {
+                    (tabster as unknown as Types.TabsterInternal).observedElement!!!.waitElement(name, timeout, accessible).then(value => {
```
ling1726, line 692: triple exclamation mark shouldn't be necessary

```suggestion
                    (tabster as unknown as Types.TabsterInternal).observedElement!.waitElement(name, timeout, accessible).then(value => {
```

core/src/State/ObservedElement.ts
```diff
@@ -9,24 +9,28 @@ import { Subscribable } from './Subscribable';
...
     private _tabster: Types.TabsterCore;
     private _initTimer: number | undefined;
-    private _waiting: {
-        [name: string]: {
-            timer?: number,
-            promise?: Promise<HTMLElement | null>,
-            resolve?: (value: HTMLElement | null) => void,
-            reject?: () => void }
-    } = {};
+    private _waiting: { [key: string]: ObservedWaiting } = {};
```
ling1726, line 33:
```suggestion
    private _waiting: Record<string, ObservedWaiting> = {};
```

core/src/CrossOrigin.ts
mshoho: I know. It's a way we used to tell something like «this is a very fragile operator, do not use it at all, but if you use it, make completely sure it's legit».

core/src/CrossOrigin.ts
```diff
@@ -347,6 +347,7 @@ interface CrossOriginElementDataIn {
     rootId?: string;
     ownerId?: string;
     observedName?: string;
+    accessible?: boolean;
```
ling1726, line 350: Not very clear what `accessible` is supposed to represent here, let's ad jsdoc and consider a rename to smth that makes more sense

core/src/State/ObservedElement.ts
```diff
@@ -270,17 +281,41 @@ export class ObservedElementAPI
...
-        const w = name && this._waiting[name];
 
-        if (w) {
-            if (w.timer) {
-                this._win().clearTimeout(w.timer);
+        if (name) {
+            const key1 = '_' + name;
+            const key2 = 'a' + name;
+            const w1 = this._waiting[key1];
+            const w2 = this._waiting[key2];
```
ling1726, line 289: let's make this variable naming clearer

ling1726 (approved):

PR #43, opened 2021-06-16
feat(version): Embed the version of the package in each tabster instance

ling1726:
Uses the rollup replace plugin to add the current version to the code
during build. This version is assigned to each tabster instance

### DIFF
--- a/core/src/Tabster.ts
+++ b/core/src/Tabster.ts
@@ -34,6 +34,7 @@ class Tabster implements Types.TabsterCore, Types.TabsterInternal {
     private _win: WindowWithTabsterInstance | undefined;
     private _forgetMemorizedTimer: number | undefined;
     private _forgetMemorizedElements: HTMLElement[] = [];
+    public _version: string | undefined;
 
     keyboardNavigation: Types.KeyboardNavigationState;
     focusedElement: Types.FocusedElementState;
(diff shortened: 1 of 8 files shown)


core/tsdx.config.js
```diff
@@ -17,7 +18,8 @@ module.exports = {
         // A workaround for babel-plugin-transform-async-to-promises used by tsdx
         // injecting helper code with `const` and producing invalid ES5 code.
         config.plugins.push(replace({
-            'const ': 'var '
+            'const ': 'var ',
+            'LOCAL_VERSION': version,
```
mshoho, line 22: This seems to be a thing inside of a hack. Ideally, this hack shouldn't even be there.
It would probably be less hacky to do version the same way __DEV__ works. It'll be some globally declared variable available anywhere.

ling1726: but having __DEV__ relies on some bundler injecting that during the build process to all usages, the only difference here is that there is no need for a global type declaration.

But I can go the way of using a global variable too and replace at build

core/src/Tabster.ts
```diff
@@ -34,6 +34,7 @@ class Tabster implements Types.TabsterCore, Types.TabsterInternal {
     private _win: WindowWithTabsterInstance | undefined;
     private _forgetMemorizedTimer: number | undefined;
     private _forgetMemorizedElements: HTMLElement[] = [];
+    public _version: string | undefined = __VERSION__;
```
mshoho, line 37: Any specific reason to have | undefined? I suppose it should always be defined.
Also I suppose all our WebPack watchers (storybook and test container) should be adjusted to process this variable.

ling1726: not really, it's a good idea to update all our webpacks with this

PR #48, opened 2021-07-13
Disabling dummy inputs tabbability when the uncontrolled element has focus.

mshoho:


core/src/__tests__/Uncontrolled.test.tsx
```diff
@@ -67,4 +67,63 @@ describe('Uncontrolled', () => {
                 expect(el?.textContent).toContain('Button4');
             });
     });
+
+    it('should allow to go outside of the application even if the uncontrolled element is the first/last in the application', async () => {
```
ling1726, line 71: Let's split this into two test cases so it's easier to maintain/read:
* first in app
* last in app

ling1726 (approved):

PR #63, opened 2021-10-07
chore(Mover): MoverAPI now handles `updateVisible`

### LINKED ISSUE #62: Mover: do not store all mover state in a static variable
### Problem

Currently mover stores all instances of itself as a static class member

https://github.com/microsoft/tabster/blob/2def22548b23906f4847c6bfd82fafd602eeebae/core/src/Mover.ts#L33

This is bad since the `dispose` of the actual `Mover` instance is only called when a DOM change is detected by the mutation observer. In scenarios where there is a child window, closing the window will not trigger mutation observer events and dispose is never called.

Therefore all `Mover` instances are retained int he static member.

### Solution 

Refactor `Mover` and `MoverAPI` so that the collection of `Mover` instances is managed by `MoverAPI` so it can be safely disposed by consuming apps, by disposing APIs through `disposeTabster`.

This will mean that `updateVisible` will need to be hoisted from `Mover` to `MoverAPI`

ling1726:
Fixes #62

Removes the static `Mover._movers` collection to be owned by the
`MoverAPI`. This means that `updateVisible` needs to be done in the
`MoverAPI`

### DIFF
--- a/core/src/Mover.ts
+++ b/core/src/Mover.ts
@@ -477,10 +477,6 @@ export class MoverAPI implements Types.MoverAPI {
             validateMoverProps(props);
         }
 
-        if (!tabster.mover) {
-            throw new Error('Mover needs to be created with getMover before being used');
-        }
-
         const self = tabster.mover as MoverAPI;
         const newMover = new Mover(tabster, element, self._onMoverDispose, props);
         self._movers[newMover.id] = newMover;


core/src/Mover.ts
```diff
@@ -488,7 +477,18 @@ export class MoverAPI implements Types.MoverAPI {
             validateMoverProps(props);
         }
 
-        return new Mover(tabster, element, props);
+        if (!tabster.mover) {
+            throw new Error('Mover needs to be created with getMover before being used');
```
mshoho, line 481: This message is already present in Instance.ts, no need to duplicate it.

PR #64, opened 2021-10-28
implement code path with dummy inputs

ling1726:
Abstracts the dummy input logic to a `DummyInputManager` class that can be reused.

The `DummyInputManager` automatically adds dummy focus sentinels as first and last child of an element and will be used in:
* Mover
* Groupper
* Modalizer

So that `FocusedElementState` no longer needs to control **every single tab key press under Root**. This makes it easier for users to adopt tabster gradually.

The new dummy input code path can be configured when creating tabster. **The current codepath is still the default**, where each tab key press is controlled

```tsx
createTabster(window, { controlTab: false });
```

Both codepaths can now be tested in the repo:

```
// Inserts dummy inputs and stops controlling tab
UNCONTROLLED=true npm test

// current main codepath
npm test
```

### DIFF
--- a/core/src/Groupper.ts
+++ b/core/src/Groupper.ts
@@ -37,7 +37,7 @@ class GroupperDummyManager extends DummyInputManager {
 export class Groupper extends TabsterPart<Types.GroupperProps> implements Types.Groupper {
     private _shouldTabInside = false;
     private _first: WeakHTMLElement | undefined;
-    public _dummyManager?: GroupperDummyManager;
+    _dummyManager?: GroupperDummyManager;
 
     constructor(
         tabster: Types.TabsterInternal,
--- a/core/src/Groupper.ts
+++ b/core/src/Groupper.ts
@@ -10,11 +10,39 @@ import { getTabsterOnElement } from './Instance';
 import { Keys } from './Keys';
 import { RootAPI } from './Root';
 import * as Types from './Types';
-import { getLastChild, TabsterPart, WeakHTMLElement } from './Utils';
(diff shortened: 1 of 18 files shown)


core/src/Groupper.ts
```diff
@@ -10,11 +10,34 @@ import { getTabsterOnElement } from './Instance';
...
+                this._tabster.focusedElement.focusLast({ container });
+            }
+        }
+    }
+}
 
 export class Groupper extends TabsterPart<Types.GroupperProps> implements Types.Groupper {
     private _shouldTabInside = false;
     private _first: WeakHTMLElement | undefined;
+    public _dummyManager?: GroupperDummyManager;
```
mshoho, line 40: `public` is not used explicitly over the codebase, perhaps we should keep it consistent here and in other changes from this PR (and have a prettier rule to use it or not once we have it).

core/src/__tests__/Focusable.test.tsx
```diff
@@ -6,7 +6,7 @@
 import * as BroTest from '../../testing/BroTest';
 import { getTabsterAttribute } from '../Tabster';
 
-describe('Focusable', () => {
+xdescribe('Focusable', () => {
```
mshoho, line 9: Why is it disabled?

ling1726: that shouldn't be, I'll remove this

core/src/Groupper.ts
ling1726: removed usages of `public` from this PR

PR #72, opened 2021-11-11
chore: add tslib to deps of tabster

layershifter:
Currently in `dist` there are imports of `tslib`, but it's not declared as a dependency:

image
https://cdn.jsdelivr.net/npm/tabster@1.0.6/dist/tabster.esm.js

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -24,7 +24,8 @@
     "test": "jest"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


ling1726:
@mshoho a dep update like this will have to be released, I think patch version, since it rather a bugfix IMO

PR #73, opened 2021-11-11
fix(keyboardNavigationState): use `nativeFocus` when simulating keyboard

ling1726:
The `nativeFocus` is the original focus before keybord hijacks it, but
also serves a special use case in the library.

Using `nativeFocus` will allow keyborg to detect that we are trying to
simulate keyboard behaviour, which will change the value of the keyboard
navigation state.

This unknown aspect was ignored in #64 and regressed, where Mover and
Groupper APIs were not updating keyboard navigation state correctly

### DIFF
--- a/core/src/Groupper.ts
+++ b/core/src/Groupper.ts
@@ -25,12 +25,11 @@ class GroupperDummyManager extends DummyInputManager {
     private _onFocusDummyInput = (dummyInput: DummyInput) => {
         const container = this._element.get();
         if (container && !dummyInput.shouldMoveOut) {
-            const findFn = dummyInput.isFirst ? 'findFirst' : 'findLast';
-            const toFocus = this._tabster.focusable[findFn]({ container });
-
-            if (toFocus) {
-                nativeFocus(toFocus);
-            }
+           if (dummyInput.isFirst) {
+                this._tabster.focusedElement.focusFirst({ container });
+            } else {
+                this._tabster.focusedElement.focusLast({ container });
+            } 
         }
     }
 }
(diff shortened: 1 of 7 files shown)


core/src/Root.ts
```diff
@@ -72,7 +72,7 @@ class RootDummyManager extends DummyInputManager {
 
                 if (hasFocused) {
                     return;
-                }
+                }
```
ling1726, line 75:
```suggestion
                }
```

core/src/Groupper.ts
```diff
@@ -25,11 +25,11 @@ class GroupperDummyManager extends DummyInputManager {
     private _onFocusDummyInput = (dummyInput: DummyInput) => {
         const container = this._element.get();
         if (container && !dummyInput.shouldMoveOut) {
-            if (dummyInput.isFirst) {
+           if (dummyInput.isFirst) {
                 this._tabster.focusedElement.focusFirst({ container });
             } else {
                 this._tabster.focusedElement.focusLast({ container });
-            }
+            }
```
ling1726, line 32:
```suggestion
            }
```

core/src/Groupper.ts
```diff
@@ -25,7 +25,7 @@ class GroupperDummyManager extends DummyInputManager {
     private _onFocusDummyInput = (dummyInput: DummyInput) => {
         const container = this._element.get();
         if (container && !dummyInput.shouldMoveOut) {
-            if (dummyInput.isFirst) {
+           if (dummyInput.isFirst) {
```
ling1726, line 28:
```suggestion
            if (dummyInput.isFirst) {
```

PR #84, opened 2021-11-22
Adds bundle-size tooling

### LINKED ISSUE #83: Add Bundle size metrics
As bundle size is critical for this project, we should add bundle size metrics and reports to keep a better track on this. The [bundle-size](https://github.com/microsoft/fluentui/tree/master/packages/bundle-size) used internally on FluentUI should be enough.

bsunderhus:


package.json
```diff
@@ -27,19 +27,22 @@
...
     },
     "devDependencies": {
+        "@azure/data-tables": "^13.0.0",
         "@babel/core": "^7.16.0",
         "@babel/plugin-proposal-class-properties": "^7.16.0",
         "@babel/plugin-transform-react-jsx": "^7.12.7",
         "@babel/plugin-transform-typescript": "^7.16.0",
         "@babel/preset-react": "^7.12.7",
         "@babel/preset-typescript": "^7.12.7",
+        "@fluentui/bundle-size": "^1.1.2",
```
ling1726, line 45: These extra dev deps should be removed once microsoft/fluentui#20612 is release right ?

bsunderhus: Yes, this is temporary

bundle-size/Tabster.fixture.js
```diff
@@ -0,0 +1,45 @@
...
+    getCurrentTabster,
+    getInternal,
+    overrideBasics,
+    makeNoOp,
+    isNoOp,
+    Types
+);
+
+export default {
+    name: "tabster",
```
layershifter, line 44: Is this fixture useful? Is there sense to have something more granular?

bundle-size/Tabster.fixture.js
```diff
@@ -0,0 +1,45 @@
...
+    getCrossOrigin,
+    getDeloser,
+    getModalizer,
+    getObservedElement,
+    getOutline,
+    getCurrentTabster,
+    getInternal,
+    overrideBasics,
+    makeNoOp,
+    isNoOp,
```
ling1726, line 18:
```suggestion
    createTabster,
    disposeTabster,
    getTabsterAttribute,
    setTabsterAttribute,
    getGroupper,
    getMover,
    getDeloser,
    getModalizer,
    getCurrentTabster,
```

I would start with this, it' s what is intended to be widely used by our partners

package.json
```diff
@@ -27,19 +27,22 @@
         "prepublishOnly": "npm run build",
         "serve": "npx http-serve storybook-static",
         "start": "npm run storybook",
-        "storybook": "start-storybook -p 8080"
+        "storybook": "start-storybook -p 8080",
+        "bundle-size": "npm run build && bundle-size measure"
```
ling1726, line 31: nit keep the ordering of scripts alphabetical

package.json
ling1726: Downgrading this version to the one before azure-tables integration would be good enough as a temp workaround to #83

bundle-size/Tabster.fixture.js
bsunderhus: instead, I'll create another fixture, It would still be good to have one for all we export IMO

bundle-size/Tabster.fixture.js
bsunderhus: I guess having one fixture with all that's been exported from tabster is good to know the actual total bundle size. I'll add a more granular one also

ling1726 (approved):

PR #91, opened 2021-11-25
fix: eslint should ignore js files

ling1726:
Doesn't impact CI, but removes annoying errors on the IDE even though the eslint command is not configured to lint anything other than `src`


ling1726:
closing since #92 will add it

PR #92, opened 2021-11-25
Flatten project roots using TS project references

bsunderhus:
Flatten project roots into 3 using TS project references

- `src`
- `tests`
- `stories`


.eslintrc.json
```diff
@@ -9,8 +9,14 @@
         "prettier"
     ],
     "parser": "@typescript-eslint/parser",
+    "ignorePatterns": "*.config.js",
     "parserOptions": {
-        "project": "tsconfig.json",
+        "projects": [
+            "src/tsconfig.json",
+            "tests/tsconfig.json",
+            "stories/tsconfig.json",
```
ling1726, line 17: Let's make this as clear as possible, since this it took me longer than I thought it would reviewing the rollup config to know what you were building there
```suggestion
            "src/tsconfig.lib.json",
            "tests/tsconfig..spec.json",
            "stories/tsconfig.stories.json",
```

ling1726: then we can name tsconfig.base => tsconfig.json

ling1726: although i think currently if you keep them all as `tsconfig.json` then vscode will correctly typecheck each file with the relevant tsconfig right ?

bsunderhus: we can't name `tsconfig.base.json` -> `tsconfig.json`, because there's already a `tsconfig.json` at the root which aggregates all the other `tsconfig.json` files which is the one used by vscode for intellisense and etc.

Regarding the names, it makes no difference actually, `tsconfig.json` on the root is the one being used by vscode if I'm not mistaken

ling1726: Oh OK I got it, let's rename the specific tsconfigs more appropriately

ling1726 (approved):

PR #94, opened 2021-11-26
chore: IE11 EOL and un-rollup files

ling1726:
Using `bundle-size` we recently discovered that due to es6 classes and
statics, things were leaking into the bundle that shouldn't be there.

This PR modifies the build to create two separate directories:

* lib -> es6 modules
* lib-commonjs -> commonjs

The PR also changes the target to ES2019 (EOL for IE11 🪦). However, this results in bundle size improvements shown below:

Before:
image

After:
image


package.json
```diff
@@ -5,9 +5,9 @@
     "author": "Marat Abdullin <marata@microsoft.com>",
     "license": "MIT",
     "sideEffects": false,
-    "main": "./dist/index.js",
-    "module": "./dist/tabster.esm.js",
-    "typings": "./dist/index.d.ts",
+    "main": "./lib-commonjs/index.js",
+    "module": "./lib/tabster.esm.js",
```
ling1726, line 9:
```suggestion
    "module": "./lib/index.js",
```

ling1726:
closed in favour of #95 to release alpha, so the repo can be tagged

PR #95, opened 2021-12-01
chore: IE11 EOL and un-rollup files

ling1726:
Using `bundle-size` we recently discovered that due to es6 classes and
statics, things were leaking into the bundle that shouldn't be there.

This PR modifies the build to create two separate directories:

* lib -> es6 modules
* lib-commonjs -> commonjs

The PR also changes the target to ES2019 (EOL for IE11 🪦). However, this results in bundle size improvements shown below:

Before:
image

After:
image


ling1726:
closed in favour of #97

PR #136, opened 2022-04-19
Create lightweight Tabster instance backed by a global TabsterCore

### LINKED ISSUE #127: Tabster should be used from a wrapper that can be disposed at any time
Similar to keyborg, the actual tabster instance should be a browser singleton. All usages of tabster should be proxies that can be disposed at any time.

The main tabster singleton should only be disposed once the last proxy wrapper is disposed

ling1726:
Follows a similar concept to keyborg. Users will use a lightweight Tabster instance while the core functionality will be private inside a global TabsterCore instance.

This means that for creating a disposing end-user Tabster instances will be cheap and safe. The global TabsterCore will only be disposed once the last Tabster instance is disposed. This is important for frameworks like react which can potentially create a dispose multiple Tabster instances in a lifecycle

Fixes #127
Fixes partially microsoft/fluentui#22291


src/Outline.ts
```diff
@@ -24,25 +24,6 @@ const defaultProps: Types.OutlineProps = {
 
 let _props: Types.OutlineProps = defaultProps;
 
-let _fullScreenEventName: string | undefined;
```
ling1726, line 27: this was a side effect and leaking into the bundle even when outline wasn't used

mshoho (approved):

PR #143, opened 2022-05-02
use internal class functions instead of static functions

ling1726:
Since we now strip `@internal` types from the final API, `dispose` and other internal public functions won't be visible to end users. We can remove most static functions from APIs

### DIFF
--- a/src/CrossOrigin.ts
+++ b/src/CrossOrigin.ts
@@ -1489,12 +1489,8 @@ export class CrossOriginFocusedElementState
         this._transactions = transactions;
     }
 
-    protected dispose() {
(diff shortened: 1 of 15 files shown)


src/CrossOrigin.ts
```diff
@@ -1489,14 +1489,10 @@ export class CrossOriginFocusedElementState
         this._transactions = transactions;
     }
 
-    protected dispose() {
+    dispose() {
         super.dispose();
```
mshoho, line 1493: I suppose we can ditch the methods which only call the superclass here and below.

mshoho (approved):

PR #146, opened 2022-05-09
feat: Allow circular focus trap in Modalizer

### LINKED ISSUE #122: Modalizer should allow focus to be completely trapped
Currently modalizer focus trap, allows the focus to escape the document into the window. This behaviour should be configurable and allow the focus trap to be circular within the modalizer and never escape

ling1726:
Implements a new prop for Modalizer `isTrapped` which allows focus to
stay within the Modalizer and never leave the it. The prop is
implemented for both controlled an uncontrolled code paths in Tabster.

Fixes #122

### DIFF
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -606,7 +606,7 @@ export interface ModalizerProps {
     isNoFocusFirst?: boolean;
     isNoFocusDefault?: boolean;
     /** A focus trap variant, keeps focus inside the modal when tabbing */
-    circular?: boolean;
+    isTrapped?: boolean;
 }
 
 export type ModalizerEventDetails = { eventName: "beforefocusout" };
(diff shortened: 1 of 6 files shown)


src/Types.ts
```diff
@@ -605,6 +605,8 @@ export interface ModalizerProps {
     isAlwaysAccessible?: boolean;
     isNoFocusFirst?: boolean;
     isNoFocusDefault?: boolean;
+    /** A focus trap variant, keeps focus inside the modal when tabbing */
+    circular?: boolean;
```
mshoho, line 609: I think we need a more speaking name here. Like trapFocus.

mshoho: or isFocusTrapped (to be like the other props here).

ling1726: named to `isTrapped`

mshoho (approved):

PR #154, opened 2022-06-15
Supporting iframes and improving the testing framework.

mshoho:


tests/utils/runner.js
```diff
@@ -0,0 +1,60 @@
...
+    if (match) {
+        let port = parseInt(match[1], 10);
+
+        if (!(port > 0)) {
+            port = 8080;
+        }
+
+        console.log(`Serving test page on port ${port}`);
+
+        testRun = spawn("jest", process.argv.slice(2), {
```
ling1726, line 41: what you're doing here is creating a custom bit of magic on top of jest, which is generally an antipattern. I would suggest creating a custom jest environment for tabster, which would be far more intuitive as a reusable thing.

https://jestjs.io/docs/configuration#testenvironment-string

mshoho: Sure, I'll look into that. This custom bit of magic is a quick way to test if the idea works in general.

PR #164, opened 2022-07-08
BREAK: Re-release [#136](https://github.com/microsoft/tabster/pull/136) and [#149](https://github.com/microsoft/tabster/pull/149) as breaking changes

mshoho:
After releasing breaking changes in minor version and after attempts to make these changes less breaking, we've ended up reverting the breaking changes for 1.5.x and releasing them as 2.0.0. This reverts all attempts to make changes less breaking and just keeps the changes for 2.0.0.

Releases changes from #136 and #149


CHANGELOG.md
```diff
@@ -4,36 +4,15 @@ All notable changes to this project will be documented in this file. Dates are d
 
 Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
-#### [v1.5.3](https://github.com/microsoft/tabster/compare/v1.5.2...v1.5.3)
```
ling1726, line 7: revert the changelog changes please, we still want them. We just want to add a changelog entry that mentions the versions being reverted

mshoho: Theoretically, it should be automatically regenerated. Let's try.

PR #174, opened 2022-08-03
chore: adds devcontainer to gitignore

bsunderhus:


ling1726 (changes requested):
I'm not quite sure we want to (or have the capacity to) maintain this in the v-build team. @Hotell  any thoughts ?

ling1726 (approved):

PR #177, opened 2022-08-09
fix: Avoid updating dummy inputs positions while scrolling to avoid excessive reflows.

mshoho:
Calling the real dummy inputs update positions function only after the scrolling ends.

### DIFF
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -863,7 +863,7 @@ class DummyInputManagerCore {
     private _firstDummy: DummyInput | undefined;
     private _lastDummy: DummyInput | undefined;
     private _transformElements: HTMLElement[] = [];
-    private _scrollFrame: number | undefined;
+    private _scrollTimer: number | undefined;
 
     constructor(
         tabster: Types.TabsterCore,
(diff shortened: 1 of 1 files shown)


jurokapsiar (approved):

src/Utils.ts
```diff
@@ -863,7 +863,7 @@ class DummyInputManagerCore {
     private _firstDummy: DummyInput | undefined;
     private _lastDummy: DummyInput | undefined;
     private _transformElements: HTMLElement[] = [];
-    private _scrollFrame: number | undefined;
+    private _scrollTimer: number | undefined;
```
ling1726, line 866: you should initialize timers to `0` because `clearTimeout` does not throw when called with a non-existing timer. That way we will have less unnecessary checks in the code, which add to bundle-size and complexity.

mshoho: That is an additional operation. And I never call clearTimeout without `if (timer)` first.

mshoho: And `if (timer)` is a much lighter check that unconditionally calling a function.

ling1726: Perhaps we're optimizing for the wrong thing here ?  there hasn't been any customer feedback about tabster performance (in Teams or otherwise), but there has been for bundle-size 🤷.

mshoho: `timer && clearTimeout(timer)` is not worse than `timer = 0; clearTimeout(timer)`.

PR #180, opened 2022-08-12
Do not handle the FocusedElement onKeyDown when using Ctrl + Tab

BryanValverdeU:
On a native app, where the host does not own the Ctrl + Tab hotkey, Tabster is handling the ctrl+tab hotkey as Tab.

Code Change.
Check whether the event contains CtrlKey before handling the onKeyEvent

### DIFF
--- a/src/State/FocusedElement.ts
+++ b/src/State/FocusedElement.ts
@@ -426,7 +426,7 @@ export class FocusedElementState
     };
 
     private _onKeyDown = (e: KeyboardEvent): void => {
-        if (e.keyCode !== Keys.Tab) {
+        if (e.keyCode !== Keys.Tab || e.ctrlKey) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


ling1726:
This LGTM  since browsers generally use CTRL for shortcuts, could you please some tests

BryanValverdeU:
Thanks for the review @ling1726 . Added a test when pressing CtrlKey with Tab.
How often we get Tabster changes to react-tabster?

ling1726:
> Thanks for the review @ling1726 . Added a test when pressing CtrlKey with Tab. How often we get Tabster changes to react-tabster?

The cadence is kind of up to us really, we just press the button to release and propagate that back to Fluent where we press another button to do the full release there.

ling1726 (approved):

BryanValverdeU:
Oh, I see, would be nice if we can release to fluent ui/react-tabster so we can use it in our project.
Thanks @ling1726 I think I cannot complete this PR, so I think a contributor needs to complete it.

PR #183, opened 2022-08-22
fix shift+tab can't navigate outside of mover in uncontrolled mode

YuanboXue-Amber:
In this example: https://codesandbox.io/s/throbbing-sea-0ciyy7?file=/src/App.js
Tab can navigate outside of mover, but Shift+Tab cannot.
note that the issue only happens when mover element is NOT the direct child of the element with uncontrolled attribute.

Behavior after fix:
Shift+Tab can always navigate outside of the mover in uncontrolled mode.


mshoho:
I suppose we can close this one as the scenario is already fixed in master.

PR #184, opened 2022-08-25
Big find elements refactoring

mshoho:
This change addresses two issues:

1. Historically, we were adding new features without seeing the whole picture as we see it now. 
2. We weren't fully aware of how TreeWalker actually works when you walk backwards in the DOM.

These two things have lead us into a state when the implementation of finding elements is rather patchy and at this point it is hard to understand and hard to adjust for new corner cases (because whenever something new gets fixed, something old gets broken).

Plus it fixes a bunch of connected bugs (primarily the Shift+Tab ones).

### DIFF
--- a/src/Focusable.ts
+++ b/src/Focusable.ts
@@ -243,7 +243,9 @@ export class FocusableAPI implements Types.FocusableAPI {
             return null;
         }
 
-        const prepareForNextElement = (shouldContinueIfNotFound?: boolean): boolean => {
+        const prepareForNextElement = (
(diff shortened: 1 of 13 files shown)


bsunderhus:
Please add some description @mshoho 🙏🏻

mshoho:
Added description.

bsunderhus (approved):
LGTM, honestly I couldn't follow most of the modifications. It would be nice to have more tests to ensure this amount of changes doesn't affect behaviour in a harmful way.

PR #186, opened 2022-08-31
fix(Groupper): Putting dummy inputs inside the groupper container when it is <li>, <td> or <th>.

mshoho:
By default Grouppers in uncontrolled mode put dummy inputs outside of the container, which is bad HTML for elements like `<li>`, `<td>` and `<th>`. So, we handle that case putting dummy inputs inside when those elements are used.


bsunderhus (approved):
LGTM, nice to see new tests to ensure this behaviour 👍🏻

PR #189, opened 2022-09-12
feat(Groupper): Ability to delegate the role of focusable groupper container to first focusble element in groupper.

mshoho:
Generally, when Groupper is Limited or LimitedTrapFocus, the groupper container focusable (so that Enter press on that container moves focus inside the groupper and Esc press moves focus from inside the groupper to the container). We are adding `delegated` property to Groupper which will allow first focusable inside the non-focusable groupper container to behave the same way as the focusable groupper container does.

The example `<div groupper tabindex=0>...</div>` should behave the same way as `<div groupper=delegated><div tabindex=0>...</div></div>` from the Enter/Esc/Tab pressing perspective, the role of focusable container is delegated to the first focusable inside the actual container.

### DIFF
--- a/tests/MoverGroupper.test.tsx
+++ b/tests/MoverGroupper.test.tsx
@@ -969,4 +969,115 @@ describe("MoverGroupper", () => {
                 );
             });
     });
+
+    it("should handle tabbing in groupper/mover/groupper when the inner groupper container is delegated and not focusable", async () => {
+        await new BroTest.BroTest(
+            (
+                <div {...getTabsterAttribute({ root: {} })}>
+                    <button>Button1</button>
+                    <div
+                        id="outer-groupper"
+                        tabIndex={0}
+                        {...getTabsterAttribute({
+                            groupper: {
+                                tabbability:
+                                    Types.GroupperTabbabilities
+                                        .LimitedTrapFocus,
+                            },
+                        })}
+                    >
+                        <button>Button2</button>
+                        <div
+                            {...getTabsterAttribute({
+                                mover: {
+                                    visibilityAware:
+                                        Types.Visibilities.PartiallyVisible,
+                                },
+                            })}
+                        >
+                            <button>Button3</button>
+                            <div
+                                id="inner-groupper"
+                                {...getTabsterAttribute({
+                                    groupper: {
+                                        delegated: true,
+                                        tabbability:
+                                            Types.GroupperTabbabilities
+                                                .LimitedTrapFocus,
+                                    },
+                                })}
+                            >
+                                <button>Button4</button>
+                                <button>Button5</button>
+                            </div>
+                        </div>
+                        <button>Button6</button>
+                    </div>
+                    <button>Button7</button>
+                </div>
+            )
+        )
+            .pressTab()
+            .activeElement((el) => {
+                expect(el?.textContent).toEqual("Button1");
+            })
+            .pressTab()
+            .activeElement((el) => {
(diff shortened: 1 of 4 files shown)


tests/MoverGroupper.test.tsx
```diff
@@ -969,4 +969,115 @@ describe("MoverGroupper", () => {
...
+            .pressTab()
+            .activeElement((el) => {
+                expect(el?.textContent).toEqual("Button3");
+            })
+            .pressTab()
+            .activeElement((el) => {
+                expect(el?.textContent).toEqual("Button6");
+            })
+            .pressTab(true)
+            .activeElement((el) => {
```
ling1726, line 1053: - focus button 3 (outside delegated)
- tab -> focus button 6 (outside delegated)
- shift + tab -> focus button 4 (inside delegated)

Why does tab move to the next element after the delegated groupper but shift+tab move to the first element of the delegated groupper ?

tests/MoverGroupper.test.tsx
```diff
@@ -969,4 +969,115 @@ describe("MoverGroupper", () => {
...
+                        <div
+                            {...getTabsterAttribute({
+                                mover: {
+                                    visibilityAware:
+                                        Types.Visibilities.PartiallyVisible,
+                                },
+                            })}
+                        >
+                            <button>Button3</button>
+                            <div
```
ling1726, line 999: how does down arrow work for a delegated groupper while in a mover ?

tests/MoverGroupper.test.tsx
mshoho: It's mover/groupper case:

```<mover>
  <button>Button3</button>
  <groupper>
    <button>Button4</button>
    <button>Button5</button>
  </groupper>
</mover>
<button>Button6</button>```

Tab from Button3 moves you outside of the mover. Tab backwards moves you to the last element in the mover which is our groupper, inactive groupper is treated atomically and first focusable (or the groupper container if the container is focusable) is the element to be focused.

Comment (ling1726) [reply tests/MoverGroupper.test.tsx:1053]: ah yeah that makes sense from the mover's perspective

File: tests/MoverGroupper.test.tsx (hunk 2 above)
Comment (ling1726) [reply tests/MoverGroupper.test.tsx:999]: Cool, can we can we test that as well ?

File: tests/MoverGroupper.test.tsx [hunk 3]
```diff
@@ -969,4 +969,115 @@ describe("MoverGroupper", () => {
                 );
             });
     });
+
+    it("should handle tabbing in groupper/mover/groupper when the inner groupper container is delegated and not focusable", async () => {
```
ling1726, line 973: This test is massive, what if you create a separate suite of delegated groupper and split this test up?

```tsx
describe('delegated groupper', () => {
  it('tabs into delegated groupper);
  it('shift+tabs into delegated groupper);
  it('mover focuses on first element of delegated groupper);
})
```

tests/MoverGroupper.test.tsx
mshoho: I believe we do test it in various tests involving grouppers. This particular test and the test above it test it too.

tests/MoverGroupper.test.tsx
mshoho: This test has average size for all tests in this file as they test various mover/groupper combinations.
There are more related tests in Groupper.test.tsx (look for `delegated` word in the test name).

ling1726 (approved):

PR #193, opened 2022-10-05
feat(Focusable): Find focusable elements iterator.

mshoho:
Adding focusable.findIterator() to be able to iterate over focusable elements. This PR also fixes a stupid bug in finding elements in backward direction.

### DIFF
--- a/src/Focusable.ts
+++ b/src/Focusable.ts
@@ -182,6 +182,10 @@ export class FocusableAPI implements Types.FocusableAPI {
         return this._findElements(true, options) || [];
     }
 
+    findIterator(options: Types.FindIteratorProps): void {
+        this._findElements(true, options);
+    }
+
     findElement(
         options: Types.FindFocusableProps
     ): HTMLElement | null | undefined {
@@ -202,6 +206,10 @@ export class FocusableAPI implements Types.FocusableAPI {
             isBackward,
             onUncontrolled,
         } = options;
+
+        const onElement = (
+            options as Types.FindIteratorProps | { onElement: undefined }
+        ).onElement;
 
         const elements: HTMLElement[] = [];
 
@@ -258,6 +266,10 @@ export class FocusableAPI implements Types.FocusableAPI {
                     delete acceptElementState.foundElement;
                     delete acceptElementState.fromCtx;
                     acceptElementState.from = foundElement;
+
+                    if (onElement && !onElement(foundElement)) {
+                        return false;
+                    }
                 }
 
                 return !!(foundElement || shouldContinueIfNotFound);
@@ -277,14 +289,13 @@ export class FocusableAPI implements Types.FocusableAPI {
 
             if (
                 this._acceptElement(lastChild, acceptElementState) ===
-                NodeFilter.FILTER_ACCEPT
+                    NodeFilter.FILTER_ACCEPT &&
+                !prepareForNextElement(true)
             ) {
-                if (!prepareForNextElement(true)) {
-                    return elements;
-                }
-            } else {
-                walker.currentNode = lastChild;
+                return elements;
             }
+
+            walker.currentNode = lastChild;
         }
 
         let foundElement: HTMLElement | null | undefined;
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -432,7 +432,25 @@ export type FindAllProps = Pick<
     | "acceptCondition"
     | "ignoreUncontrolled"
     | "ignoreAccessibiliy"
-> & { container: HTMLElement; skipDefaultCheck?: boolean };
+> & {
+    container: HTMLElement;
+};
+
+export type FindIteratorProps = Pick<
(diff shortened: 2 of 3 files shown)


src/Focusable.ts
```diff
@@ -182,6 +182,10 @@ export class FocusableAPI implements Types.FocusableAPI {
         return this._findElements(true, options) || [];
     }
 
+    findIterator(options: Types.FindIteratorProps): void {
```
ling1726, line 185: How does this make things simpler?

```ts
let node;
while(condition(node)) {
  node = findNextFocusable(node);
}
```

Why is the above worse? This iterator pattern does not behave like any iterator pattern that I've ever seen 😨 and would be quite confusing since it relies on an internal API that isn't public either

if we really want an iterator then we should explicitly create a `FocusableIterator` that encapsulates what you want to do here with `.nextNode` `.prevNode`. which honestly feels like it could replace `findNext` and ` findPrev`

src/Focusable.ts
```diff
@@ -203,6 +207,10 @@ export class FocusableAPI implements Types.FocusableAPI {
             onUncontrolled,
         } = options;
 
+        const onElement = (
+            options as Types.FindIteratorProps | { onElement: undefined }
```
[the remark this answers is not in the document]
mshoho: _findElement is the basic internal method used by all. The callback is not optional from the iterator perspective, but from the generic methods that does it all, it's optional. If I do something like `if ((options as Types.FindIteratorProps).onElement) { ... }`, it will be a line of generated code. And here it's solved strictly on typings level and adds nothing extra to the generated code.

src/Focusable.ts
mshoho: Because `findNext` and `findPrev` start the process from scratch — they initalize the state, get context and do work to prepare for the actual search. Iterator allows to do that work once and reuse all preparation steps.

mshoho: Perhaps Iterator is the wrong name for this one. This one behave's like JavaScript array's `some()`. I can rename it to `findSome()`.

ling1726: Why does this iterator not return anything? is the only way to get elements from the `find` to use an `onElement` callback ?

mshoho: So, it's not really an iterator. It's a thing that calls callback for every element that it finds and allows to stop search by returning false. I can rename it to findSome and return the resulted array to be consistent with JavaScript's some. Would that be satisfactory?

mshoho: Basically, it's a version of `findAll` that has a callback, can start from custom element and search in any direction and can stop search when needed. I've initially done it in `findAll`, but it seemed too much for a little method finding all, so I have a new one.

mshoho: Renamed and returned the value. Does it look better now?

mshoho: Actually, it's not even `.some()` and more like `.filter()`. But in that context and with the capabilities, I suppose `.findSome()` is a reasonable name.

src/Types.ts
```diff
@@ -432,7 +432,25 @@ export type FindAllProps = Pick<
...
+    | "container"
+    | "currentElement"
+    | "isBackward"
+    | "includeProgrammaticallyFocusable"
+    | "acceptCondition"
+    | "ignoreUncontrolled"
+    | "ignoreAccessibiliy"
+> & {
+    container: HTMLElement;
+    onElement: FindElementCallback;
```
ling1726, line 450: Please add some documentation for `onElement` and `FindElementCallback` that the return boolean means that the search should stop

src/Focusable.ts
```diff
@@ -203,6 +207,10 @@ export class FocusableAPI implements Types.FocusableAPI {
             onUncontrolled,
         } = options;
 
+        const onElement = (
+            options as Types.FindSomeProps | { onElement: undefined }
```
ling1726, line 211: why can't this option be supported by all `find` APIs ?

mshoho: It makes no sense for methods that find single element (i.e. for everything but `findAll`). I can add it to `findAll`. The new method supports finding elements from a certain element forward or back. `findAll` name doesn't feel exactly correct semantically because of «all». We can remove both `findSome` and `findAll` and have just one `findElements` method (or just `find`). But that would require major version increment. Are we currently using `findAll` in FUI?

mshoho: I will move `onElement` to `FindFocusableProps`. `FindFirstProps`, `FindNextProps` and others use pick to get props from there anyway. But it still makes no sense to pick it for a single element search methods.

mshoho: Ok, I'll go back to what I was doing initially — add it all to `findAll`.

mshoho: So, it all is a part of `findAll` now.

ling1726 (approved):

PR #196, opened 2022-10-06
fix(Mover): Fixing Mover grid behaviour and adding tests for it.

mshoho:
The algorithm to calculate vertically next element in grid was insufficient. Replacing it with a better working one and adding tests.

### DIFF
--- a/src/Mover.ts
+++ b/src/Mover.ts
@@ -795,7 +795,18 @@ export class MoverAPI implements Types.MoverAPI {
             isBoth || direction === Types.MoverDirections.Horizontal;
         const isGrid = direction === Types.MoverDirections.Grid;
         const isCyclic = moverProps.cyclic;
+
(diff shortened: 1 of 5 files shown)


ling1726 (dismissed):

src/Mover.ts
```diff
@@ -614,6 +614,39 @@ function validateMoverProps(props: Types.MoverProps): void {
...
+function getDistance(
+    ax1: number,
+    ay1: number,
+    ax2: number,
+    ay2: number,
+    bx1: number,
+    by1: number,
+    bx2: number,
+    by2: number
+): number {
```
ling1726, line 639: I would suggest a nicer form for this

```ts
getDistance({x1, y1, x2, y2}, {x1, y1, x2, y2})
```

ling1726 (approved):

PR #209, opened 2022-12-08
fix: Mover should not ignore non-text inputs

### LINKED ISSUE #211: Mover does not work for checkboxes
https://codesandbox.io/s/pedantic-robinson-qklozr?file=/src/App.js

It's not possible to use arrows to navigate a group of checkboxes with the mover

ling1726:
The `Mover`current ignores all inputs by using the selector `input` https://github.com/microsoft/tabster/blob/90ca3718a81fe53730a23d6cddae87cbbf54ff2f/src/Mover.ts#L28

- updates the current `_inputSelector` to be `_textInputSelector` to be more specific and adds a specific list of exclusions under `_nonTextInputSelector`

Fixes #211

### DIFF
--- a/src/Mover.ts
+++ b/src/Mover.ts
@@ -25,7 +25,11 @@ import {
     WeakHTMLElement,
 } from "./Utils";
 
-const _inputSelector = ["input", "textarea", "*[contenteditable]"].join(", ");
+const _inputSelector = [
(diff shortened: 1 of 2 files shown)


ling1726:
Closed in favour of #212

PR #218, opened 2023-01-14
Modalizer revamp

mshoho:
* Adding ability to have composite Modalizers (one modal can be spread across different HTML nodes).
* Making uncontrolled mode better uncontrolled (Modalizers were not uncontrolled at all, Grouppers and Movers in  uncontrolled mode were not fully utilizing dummy inputs).
* Making Modalizer work nicely with the Groupper on the same node.
* All focusable element search functions require container.
* Adding some convenience DOM events.


src/Focusable.ts
```diff
@@ -129,37 +117,35 @@ export class FocusableAPI implements Types.FocusableAPI {
...
+            if (!this._tabster.modalizer?.isAugmented(el)) {
+                return true;
+            }
         }
 
         return false;
     }
 
     findFirst(options: Types.FindFirstProps): HTMLElement | null | undefined {
         return this.findElement({
```
jurokapsiar, line 129: maybe you can now just pass options - valid for each of these 3 findElement calls

src/AttributeHelpers.ts
```diff
@@ -0,0 +1,99 @@
...
+    update?: boolean
+): void {
+    let props: Types.TabsterAttributeProps | undefined;
+
+    if (update) {
+        const attr = element.getAttribute(Types.TabsterAttributeName);
+
+        if (attr) {
+            try {
+                props = JSON.parse(attr);
```
ling1726, line 76: JSON.parse is one of the slowest APIs to exist, would it be worth storing the tabster attribute config on the DOM element to avoid parsing and then stringifying JSON again?

src/AttributeHelpers.ts
```diff
@@ -0,0 +1,99 @@
...
+            } catch (e) {
+                /**/
```
ling1726, line 78: catch should at least console.warn, or even throw in development

ling1726:
The PR title will go directly into the changelog, since it's such a breaking change we should make the title more descriptive with a PR description for future reference

src/Groupper.ts
```diff
@@ -34,48 +34,36 @@ class GroupperDummyManager extends DummyInputManager {
                 relatedTarget: HTMLElement | null
             ) => {
                 const container = element.get();
+                const input = dummyInput.input;
```
ling1726, line 37: are the groupper changes even related to modalizer?

src/Root.ts
```diff
@@ -256,31 +269,60 @@ export class RootAPI implements Types.RootAPI {
...
+        } else if (!this._autoRootWaiting) {
+            this._autoRootWaiting = true;
+            doc.addEventListener("readystatechange", this._autoRootCreate);
+        }
+
+        return undefined;
+    };
+
+    private _autoRootUnwait(doc: Document): void {
+        if (this._autoRootWaiting) {
```
ling1726, line 315: you don't need this condition right? `removeEventListener` does not throw when it cannot find the event listener, and you can just always set the flag to falst

ling1726:
A list of breaking API changes in this PR would also be helpful

src/Utils.ts
```diff
@@ -949,10 +955,34 @@ interface DummyInputWrapper {
...
+            ),
+        ].join(", ")
+    );
+}
+
 /**
  * Parent class that encapsulates the behaviour of dummy inputs (focus sentinels)
  */
 class DummyInputManagerCore {
+    private _tabster: Types.TabsterCore;
```
ling1726, line 985: this tabster instance is never being de-referenced after `dispose`

src/Modalizer.ts
```diff
@@ -306,74 +317,37 @@ function validateModalizerProps(props: Types.ModalizerProps): void {
...
-    private _focusOutTimer: number | undefined;
     private _restoreModalizerFocusTimer: number | undefined;
-    /**
-     * Modalizers managed by this API, stored by id
-     */
     private _modalizers: Record<string, Types.Modalizer>;
+    private _parts: Record<string, Record<string, Types.Modalizer>>;
+    private _augMap: WeakMap<HTMLElement, true>;
+    private _aug: WeakRef<HTMLElement>[];
+    private _hiddenUpdateTimer: number | undefined;
```
ling1726, line 329: You need to clear the `_hiddenUpdateTimer` on dispose

src/Focusable.ts
mshoho: Not sure what you mean.

ling1726 (approved):

PR #222, opened 2023-01-24
fix(Mover): Correct grid behaviour for Home/End PgUp/PgDn

ling1726:
Fixes the mover grid behaviour for Home/End so that the keys will focus the first and last elements in the row respectively.

PageUp and PageDown keys should preserve focused column per WAI best practices https://www.w3.org/WAI/ARIA/apg/example-index/grid/dataGrids

Also updates `isElementVerticallyVisibleInContainer` to round bounding rect values to integers in the same way as it is done with the `Mover`.

* top - Math.ceil
* bottom - Math.floor


src/Mover.ts
```diff
@@ -895,13 +895,64 @@ export class MoverAPI implements Types.MoverAPI {
                 ignoreUncontrolled: true,
                 useActiveModalizer: true,
             });
+            if (isGrid) {
+                // finds the last focusable element on the horizontal
+                next = undefined;
+                let cur: HTMLElement | undefined | null = focusable.findPrev({
```
mshoho, line 901: I suppose we can use `findElement()` with `acceptCondition` and avoid the loop — check the `acceptElement` for `visibilityAware` condition above.

src/Mover.ts
```diff
@@ -895,13 +895,64 @@ export class MoverAPI implements Types.MoverAPI {
...
+                    container,
+                    currentElement: focused,
+                    ignoreUncontrolled: true,
+                    useActiveModalizer: true,
+                });
+                let nextElementX2 = Math.ceil(
+                    cur?.getBoundingClientRect().right ?? 0
+                );
+
+                while (cur && nextElementX2 > focusedElementX1) {
```
mshoho, line 940: Same here for `findElement()`.

src/Mover.ts
```diff
@@ -895,13 +895,64 @@ export class MoverAPI implements Types.MoverAPI {
...
+                        container,
+                        currentElement: cur,
+                    });
+                    nextElementX1 = Math.ceil(
+                        cur?.getBoundingClientRect().left ?? -1
+                    );
+                }
+            }
         } else if (keyCode === Keys.End) {
             next = focusable.findLast({
```
mshoho, line 922: Let's do `if (isGrid) {} else {}` to avoid unnecessary `findLast()` call. It is a few more symbols, but that function call is not the most lightweight thing in the world.

src/Mover.ts
ling1726: @mshoho I've updated to use findElement, can you take a look?

src/Mover.ts
```diff
@@ -863,9 +863,53 @@ export class MoverAPI implements Types.MoverAPI {
...
+                        currentElement: cur,
+                    });
+                    nextElementX2 = Math.ceil(
+                        cur?.getBoundingClientRect().left ?? 0
+                    );
+                }
+            }
         } else if (keyCode === Keys.PageUp) {
+            if (isGrid) {
+                return;
```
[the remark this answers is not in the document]
ling1726: Talked offline, I'll fix the page up/down behaviour here

mshoho (approved):

PR #229, opened 2023-02-02
feat: implement `visibilityTolerance` prop to Mover

ling1726:
Adds a tolerance of 80% to the check `isElementVerticallyVisible`. This means that when an element is cut off by the viewport, as long as the part cut off is 80% visible, the check will return true.

The Mover now supports a `visibilityTolerance` prop so that this tolerance is configurable for each mover. Adds tests to make sure that this tolerance is correctly enforced in scroll containers where elements are partially visible.

This is needed since browser scrolling is not precise enough to have 100% of an element visible all the time. Previously the tolerance was essentially 1px, which is too small.

### DIFF
--- a/src/Mover.ts
+++ b/src/Mover.ts
@@ -107,6 +107,7 @@ export class Mover
     private _updateQueue: MoverUpdateQueueItem[] | undefined;
     private _updateTimer: number | undefined;
 
+    visibilityTolerance: number;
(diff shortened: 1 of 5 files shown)


mshoho:
This is actually a good candidate to be a mover prop. It can be optional with 90% by default, but makes sense to make it adjustable.

mshoho (approved):

PR #231, opened 2023-02-02
Fixing handling various combinations of nested grouppers and movers.

mshoho:
* Fix for handling multiple levels of nestedness of movers and grouppers.
* Fix for uncontrolled mode to properly move focus when the next element to focus is out of the DOM order.


CHANGELOG.md
```diff
@@ -4,10 +4,19 @@ All notable changes to this project will be documented in this file. Dates are d
 
 Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
```
ling1726, line 5: revert the changelog and package.json changes

src/State/FocusedElement.ts
```diff
@@ -409,35 +409,55 @@ export class FocusedElementState
         }, 0);
 
         const callFindNext = (
```
ling1726, line 411: Please ass some comments what this fn is doing, it went from being a proxy call to potentially a recursive call based on mover/groupper position

PR #237, opened 2023-02-16
fix: Add ArrowLeft and ArrowRight to ignoreKeydown type

george-cz:
The library in fact supports this configuration, but the FocusableProps.ignoreKeydown doesn't include it for some reason.

This is needed for my POC of keyboard navigation in the table and interacting with an intractable element:

https://github.com/microsoft/fluentui/pull/26873/files#diff-91a26c39f89f4f6f6ce48e2d580e0c68d2b10706c96f851e5a070a65f0871fe7R47

### DIFF
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -371,6 +371,8 @@ export interface FocusableProps {
         Enter?: boolean;
         ArrowUp?: boolean;
         ArrowDown?: boolean;
+        ArrowLeft?: boolean;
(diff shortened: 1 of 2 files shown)


ling1726 (approved):

ling1726:
Can you add tests here? https://github.com/microsoft/tabster/blob/c1b46b478fd0d75fe95b5653d525658e2bf16d6a/tests/Mover.test.tsx#L327

PR #239, opened 2023-02-23
feat(sys): System API for adjusting the dummy inputs position in rare cases of necessity.

mshoho:
In uncontrolled mode we redirect focus using dummy inputs (invisible focusable elements whose only purpose is to move focus to the proper element once focused). These dummy inputs observe their position in DOM and adjust it when the application changes its DOM. Depending on the use case the dummy inputs can be positioned inside the container element or outside of it.

For example, if we have Mover on a `<div>`, the dummy inputs will be created as first and last element in the `<div>`, but if we have Mover on a `<table>`, the dummy inputs will be created outside (because having them inside is invalid HTML and causes glitches).

Generally the placement of the dummy inputs is not the application creator's concern. However, there are corner cases when we want to change the exact placement. One of the examples is when we have Mover on a `<div>` and the application keeps updating the DOM inside that `<div>`, Tabster will reposition the dummy inputs after every DOM update (to make sure that the dummy inputs are first and last in the `<div>`). Depending on application structure and logic, it might be a performance concern (because changing the DOM causes the reflow) and putting the dummy inputs outside of the `<div>` might be a way to mitigate the problem.

We are adding `sys` API for the internal Tabster's logic adjustments like this. It's not a part of Mover, Groupper and Modalizer because it doesn't belong there semantically and because of possible conflicting declarations when, for example, Groupper and Modalizer are used on the same element.

Another change is the adjustment of how the `data-tabster` properties are passed to the constructors internally (which doesn't affect public API, but makes the pull request a bit bigger). Now the entire set of props is available inside the constructors, not just the relevant to the constructor part.


src/Instance.ts
```diff
@@ -121,14 +122,14 @@ export function updateTabsterByAttribute(
...
-                        newTabsterProps.deloser as Types.DeloserProps
+                        newTabsterProps as Types.TabsterAttributePropsWith<"deloser">
                     );
                 } else {
                     if (tabster.deloser) {
                         tabsterOnElement.deloser =
                             tabster.deloser.createDeloser(
                                 element,
-                                newTabsterProps.deloser as Types.DeloserProps
+                                newTabsterProps as Types.TabsterAttributePropsWith<"deloser">
```
ling1726, line 132: There is no need to pass all props to tabster API, if all we need to use is the Sys API, we can just add that as an argument to the constructor.

This breaks down encapsulation even more and creates really strange utility type to have to maintain

src/Types.ts
```diff
@@ -776,6 +785,23 @@ export type RootConstructor = (
...
+export interface SysProps {
+    /**
+     * Force dummy input position outside or inside of the element.
+     * By default (when undefined), the position is determined dynamically
+     * (for example inside for <li> elements and outside for <table> elements,
+     * plus a default Groupper/Mover/Modalizer implementation position).
+     * Setting to true will force the dummy inputs to be always outside of the element,
+     * setting to false will force the dummy inputs to be always inside.
+     */
+    dummyInputsOutside?: boolean;
```
ling1726, line 802: This should be an enum, having Boolean properties that do something different on undefined kind of smells

src/Utils.ts
```diff
@@ -1048,13 +1060,16 @@ class DummyInputManagerCore {
 
         // Some elements allow only specific types of direct descendants and we need to
         // put our dummy inputs inside or outside of the element accordingly.
-        const tagName = element.get()?.tagName;
+        const forcedDummyPosition = sys?.dummyInputsOutside;
```
ling1726, line 1063: Something like this would be cleaner

```ts
switch(tagName) {
  case 'UL':
  case 'OL':
  case 'Table':
    this._isOutside = true;
    break;
  default:
    this._isOutside = false;
}

this._isOutside = forcedDummyPosition ?? this._isOutside;

```

ling1726 (changes requested):
I don't think maintaining a utility type and passing all props to all APIs is a good way forward, it will make encapsulation harder in the future

src/Instance.ts
mshoho: I wouldn't call it «breaking the encapsulation». We do separate logic for the particular things, but they still influence each other when nested (like, for example, a groupper state influences the upper mover behaviour). Up to that point we just didn't need to know this influence at the constructor level — we could get it all from the stored props. But now we need it in the constructor, before the props are stored. We set data-tabster for the element, I don't think it's wrong to be able to poke into it in the constructors (as we already do it after the constructors).

src/Utils.ts
mshoho: I will adjust it, but it is much less bundle size friendly :).

mshoho: Actually, the condition is more complex than that and rewriting it to switch/cases makes it more monstrous...

src/Types.ts
mshoho: I've made it a Tabster-style enum (which I hate but it's much better from the final bundle perspective comparing to what TypeScirpt does to enums).

src/Instance.ts
mshoho: Another example of the influence closer to this example is Modalizer and Groupper on the same node. The presence of one affects another one.

mshoho: And it seems to be more future-proof to me. I can pass `sys` to every constructor. But then we will add some `supercool` that's required in the constructor and will be forced to pass it down the constructors as well.

mshoho: One more example from the HTML realm. I would compare entire Tabster with `<table>`. Table is a composite entity that consists of a set of intertwined tags and attributes to achieve a layout goal. Tabster is also a set of intertwined things achieve a focus behaviour goal. One part of the table can affect another one and the parts could be correctly or incorrectly composed, like `colspan` and `rowspan` attributes affect the way cells are handled. `<tr>`, `<th>` and `<td>` are different separate parts for different things, but an attribute on one of them might affect each other in different directions. So, the Tabster itself is an encapsulated from the app part, and it consists of the components that are separated from each other code-wise to contain the logic of a particular component and make them maintainable. But they are not isolated components, they serve the singular focus management goal and might be dependent on the composition.

ling1726 (approved):

PR #249, opened 2023-03-08
fix(Modalizer): Fixing the race condition in Modalizer dispose

mshoho:


ling1726 (changes requested):
There's no need to have a `isDsposed` flag if you're not using it anywhere else. You could just dispose all the modalizers before clearing the timeout

ling1726 (approved):

PR #251, opened 2023-03-09
fix(Tabster): Initializing Tabster-related everything for the DOM elements mounted before Tabster creation.

mshoho:
Making sure that if the DOM with `data-tabster` attributes is created before Tabster, everything will be initialized properly.

### DIFF
--- a/src/Tabster.ts
+++ b/src/Tabster.ts
@@ -59,6 +59,7 @@ class TabsterCore implements Types.TabsterCore {
     private _forgetMemorizedTimer: number | undefined;
     private _forgetMemorizedElements: HTMLElement[] = [];
     private _wrappers: Set<Tabster> = new Set<Tabster>();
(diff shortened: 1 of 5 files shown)


src/Tabster.ts
```diff
@@ -117,9 +118,12 @@ class TabsterCore implements Types.TabsterCore {
             },
         };
 
-        this.internal.resumeObserver(false);
-
         startFakeWeakRefsCleanup(getWindow);
+
+        this._initTimer = win.setTimeout(() => {
```
ling1726, line 123:
```suggestion
        // Gives a tick to the host app to initialize other tabster
        // APIs before tabster starts observing attributes
        this._initTimer = win.setTimeout(() => {
```

ling1726 (approved):
Would be nice to add a comment about why the mysterious timeout is there (it's not obvious unlesss you consider things like React)

PR #254, opened 2023-03-22
feat(DummyInput): Using only main mutation observer to update dummy inputs in DOM and bulk-updating transform offsets

mshoho:
Two optimizations regarding dummy inputs:

1. There is no mutation observed in `DummyInputManagerCore` anymore. We're utilizing the main mutation observer to keep the dummy inputs DOM placement consistent. And the DOM placement update is throttled to reduce the amount of reflows.

2. Every `DummyInputManagerCore` doesn't update the dummy input visual position on scroll independently, it is being queued to do the bulk-update in order to reduce the amount of reflows.

### DIFF
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -1006,14 +1006,10 @@ export class DummyInputObserver implements Types.DummyInputObserver {
 
     remove(dummy: HTMLElement): void {
         const dummyInputElements = this._dummies;
-        const cb = dummyInputElements.get(dummy);
+        dummyInputElements.delete(dummy);
 
-        if (cb) {
-            dummyInputElements.delete(dummy);
-
-            if (dummyInputElements.size === 0) {
-                delete this.domChanged;
-            }
+        if (dummyInputElements.size === 0) {
+            delete this.domChanged;
         }
     }
 
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -994,52 +994,48 @@ export class DummyInputObserver implements Types.DummyInputObserver {
     private _domChangedTimer?: number;
     private _dummyMap: WeakMap<HTMLElement, () => void> = new WeakMap();
     private _dummies: WeakRef<HTMLElement>[] = [];
-    domChanged?(parent: HTMLElement | null): void;
+    domChanged?(parent: HTMLElement): void;
 
     constructor(win: GetWindow) {
         this._win = win;
     }
 
-    add(dummy: HTMLElement | undefined, callback: () => void): void {
-        if (dummy) {
-            const dummyMap = this._dummyMap;
-            const cb = dummyMap.get(dummy);
+    add(dummy: HTMLElement, callback: () => void): void {
+        const dummyMap = this._dummyMap;
+        const cb = dummyMap.get(dummy);
 
-            if (!cb) {
-                this._dummies.push(new WeakRef(dummy));
-                this.domChanged = this._domChanged;
-            }
-
-            dummyMap.set(dummy, callback);
+        if (!cb) {
+            this._dummies.push(new WeakRef(dummy));
+            this.domChanged = this._domChanged;
         }
+
+        dummyMap.set(dummy, callback);
     }
 
-    remove(dummy: HTMLElement | undefined): void {
-        if (dummy) {
-            const dummyMap = this._dummyMap;
-            const cb = dummyMap.get(dummy);
+    remove(dummy: HTMLElement): void {
+        const dummyMap = this._dummyMap;
+        const cb = dummyMap.get(dummy);
 
-            if (cb) {
-                dummyMap.delete(dummy);
+        if (cb) {
+            dummyMap.delete(dummy);
 
-                const dummies = this._dummies;
+            const dummies = this._dummies;
 
-                for (let i = dummies.length; i--; ) {
-                    const el = dummies[i].deref();
+            for (let i = dummies.length; i--; ) {
+                const el = dummies[i].deref();
 
-                    if (el) {
-                        if (el === dummy) {
-                            dummies.splice(i, 1);
-                            break;
-                        }
-                    } else {
+                if (el) {
+                    if (el === dummy) {
                         dummies.splice(i, 1);
+                        break;
                     }
+                } else {
+                    dummies.splice(i, 1);
                 }
+            }
 
-                if (dummies.length === 0) {
-                    delete this.domChanged;
-                }
+            if (dummies.length === 0) {
+                delete this.domChanged;
             }
         }
     }
@@ -1064,30 +1060,31 @@ export class DummyInputObserver implements Types.DummyInputObserver {
         delete this._win;
     }
 
-    private _domChanged = (parent: HTMLElement | null): void => {
-        if (parent && !this._domChangedMap.has(parent)) {
-            this._domChangedMap.set(parent, true);
-
-            if (!this._domChangedTimer) {
-                this._domChangedTimer = this._win?.().setTimeout(() => {
-                    delete this._domChangedTimer;
-
-                    for (const dummyRef of this._dummies) {
-                        const dummy = dummyRef.deref();
-                        const dummyParent = dummy?.parentElement;
-
-                        if (
-                            dummyParent &&
-                            this._domChangedMap.has(dummyParent)
-                        ) {
-                            this._dummyMap.get(dummy)?.();
-                        }
-                    }
-
-                    this._domChangedMap = new WeakMap();
-                }, 100);
-            }
+    private _domChanged = (parent: HTMLElement): void => {
+        if (this._domChangedMap.has(parent)) {
+            return;
         }
+
+        this._domChangedMap.set(parent, true);
+
+        if (this._domChangedTimer) {
+            return;
+        }
+
+        this._domChangedTimer = this._win?.().setTimeout(() => {
+            delete this._domChangedTimer;
+
+            for (const dummyRef of this._dummies) {
+                const dummy = dummyRef.deref();
+                const dummyParent = dummy?.parentElement;
+
+                if (dummyParent && this._domChangedMap.has(dummyParent)) {
+                    this._dummyMap.get(dummy)?.();
+                }
+            }
+
+            this._domChangedMap = new WeakMap();
+        }, 100);
     };
 
     updateOffsets(callback: () => void): void {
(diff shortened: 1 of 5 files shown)


src/Utils.ts
```diff
@@ -986,12 +986,129 @@ function setDummyInputDebugValue(
...
+            const dummyMap = this._dummyMap;
+            const cb = dummyMap.get(dummy);
+
+            if (!cb) {
+                this._dummies.push(new WeakRef(dummy));
+                this.domChanged = this._domChanged;
+            }
+
+            dummyMap.set(dummy, callback);
+        }
```
bsunderhus, line 1014:
```suggestion
        if (!dummy) {
          return
        }
        const dummyMap = this._dummyMap;
        const cb = dummyMap.get(dummy);

        if (!cb) {
            this._dummies.push(new WeakRef(dummy));
            this.domChanged = this._domChanged;
        }

        dummyMap.set(dummy, callback);
```

avoid adding another level of nesting by short returning (the less nested the easier to read)

src/Utils.ts
```diff
@@ -986,12 +986,129 @@ function setDummyInputDebugValue(
...
+                    } else {
+                        dummies.splice(i, 1);
+                    }
+                }
+
+                if (dummies.length === 0) {
+                    delete this.domChanged;
+                }
+            }
+        }
```
bsunderhus, line 1044:
```suggestion
        if (!dummy) {
          return
        }
        const dummyMap = this._dummyMap;
        const cb = dummyMap.get(dummy);

        if (!cb) {
          return
        }

        dummyMap.delete(dummy);

        const dummies = this._dummies;

        for (let i = dummies.length; i--; ) {
            const el = dummies[i].deref();

            if (el === dummy) {
              dummies.splice(i, 1);
              break;
            }
            if (el === undefined) {
              dummies.splice(i, 1);
            }
        }

        if (dummies.length === 0) {
            delete this.domChanged;
        }
```

avoid adding another level of nesting by short returning (the less nested the easier to read)

src/Utils.ts
```diff
@@ -986,12 +986,129 @@ function setDummyInputDebugValue(
...
+                            this._domChangedMap.has(dummyParent)
+                        ) {
+                            this._dummyMap.get(dummy)?.();
+                        }
+                    }
+
+                    this._domChangedMap = new WeakMap();
+                }, 100);
+            }
+        }
```
bsunderhus, line 1090:
```suggestion
        if (!parent || this._domChangedMap.has(parent)) {
          return
        }
        
        this._domChangedMap.set(parent, true);

        if (this._domChangedTimer) {
          return
        }

        this._domChangedTimer = this._win?.().setTimeout(() => {
            delete this._domChangedTimer;

            for (const dummyRef of this._dummies) {
                const dummy = dummyRef.deref();
                const dummyParent = dummy?.parentElement;

                if (
                    dummyParent &&
                    this._domChangedMap.has(dummyParent)
                ) {
                    this._dummyMap.get(dummy)?.();
                }
            }

            this._domChangedMap = new WeakMap();
        }, 100);
```

avoid adding another level of nesting by short returning (the less nested the easier to read)

bsunderhus (commented):
Added a few comments to improve readability a bit!

src/Utils.ts
mshoho: `if (false) { return; } doThings();'` is less bundle-size friendly than `if (true) { doThings(); }`  — that is practically the only reason. But sure.

src/Utils.ts
```diff
@@ -986,12 +986,129 @@ function setDummyInputDebugValue(
     );
 }
 
+export class DummyInputObserver implements Types.DummyInputObserver {
+    private _win?: GetWindow;
+    private _offsetsQueue: (() => void)[] = [];
+    private _offsetsTimer?: number;
```
ling1726, line 992: I would prefer `resetDummyInputs` rather than `offset` since the current name really has no relevance to the operation that is happening

src/Utils.ts
```diff
@@ -986,12 +986,129 @@ function setDummyInputDebugValue(
...
+    private _domChangedMap: WeakMap<HTMLElement, true> = new WeakMap();
+    private _domChangedTimer?: number;
```
ling1726, line 994:
```suggestion
    private _pendingDummyInputResets: WeakSet<HTMLElement> = new WeakSet();
    private _resetDummyInputsTimer?: number;
```
I'd rename these to be clearer what they do, also if you're just mapping something to a boolean, you can use a [WeakSet](https://caniuse.com/?search=JavaScript%20WeakSet).

The same comment applies about the weak structure, there is an explicit dispose step so I don't think weak structures are necessary

src/Utils.ts
```diff
@@ -986,12 +986,129 @@ function setDummyInputDebugValue(
...
+    private _dummyMap: WeakMap<HTMLElement, () => void> = new WeakMap();
+    private _dummies: WeakRef<HTMLElement>[] = [];
```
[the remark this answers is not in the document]
mshoho: `add`/`remove` is when the particular dummy input is added to or removed from the DOM (happens multiple times during the application life cycle). `dispose` is for the entire Tabster dispose.

The entries need to be enumerated during the update. The map is to have a hash with the element as the key (JS only supports strings as keys for normal objects) for fast checks. The list is to be able to go through them (as WeakMap doesn't support enumeration). WeakRefs are everywhere in Tabster by design, because Tabster only follows the application DOM, not creates it (and bizarre things happen in the applications) — last thing you want is to be constantly chased by the people investigating memory leaks and blaming you because you hold a strong reference.

Other than that — I just think it's a good idea to make sure things are explicitly clear on dispose.

src/Utils.ts
```diff
@@ -1079,24 +1196,20 @@ class DummyInputManagerCore {
             element
         );
 
+        // We will be checking dummy input parents to see if their child list have changed.
+        // So, it is enough to have just one of the inputs observed, because
+        // both dummy inputs always have the same parent.
+        const dummyElement = this._firstDummy.input;
```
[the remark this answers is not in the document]
mshoho: I suppose that would make it much worse. From the performance perspective it will add much more background payload. Calling a function (`shouldUpdate`) for all instances on all updates is a way heavier than just calling a callback for the subset of instances (created by inline-checking for the presence in a map of changed elements).

Plus `shouldUpdate` would either have to receive the map of changed parents or check the dummy inputs current state all the time, even if they are not in the changed scope.

I try to optimize for performance here as it is constantly running background task. Instead of HTML elements I can track higher level instances. But the actual elements are needed for actual operations (and I wouldn't want to pay the price of uselessly unwrapping the higher level entities all the time in the background task).

src/Utils.ts
```diff
@@ -986,12 +986,129 @@ function setDummyInputDebugValue(
...
+
+        if (!this._offsetsTimer) {
+            this._offsetsTimer = this._win?.().setTimeout(() => {
+                delete this._offsetsTimer;
+
+                this._offsetsQueue.forEach((callback) => callback());
+                this._offsetsQueue = [];
+            }, 100);
+        }
+    }
```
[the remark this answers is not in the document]
mshoho: Well, if we speak about consistency, it would require to change quite a number of places. Because to that day I was mostly everywhere doing `if (!timer) { doThings(); }` rather than `if (timer) { return; } doThings();`. Because it's better for the bundle size. @bsunderhus has asked me to change these in _domChanged for more readability.

I suppose if we really want to be consistent about it, we can try to change it everywhere (maybe even try some lint rule, if possible). But we need to decide what is more important — this kind of consistency or a bundle size. Perhaps the JS minifiers can adjust these cases automatically, makes sense to check.

src/Utils.ts
mshoho: `_pendingChangedParents`?

With these names I also always struggle to make them shorter (as the prop names do not get minified at all).

src/Utils.ts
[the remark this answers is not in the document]
ling1726: I don't mind which one you do, you can make `_domChanged` the same as what you have here, it's just a bit strange seeing both conventions in the same class

src/Utils.ts
ling1726: > With these names I also always struggle to make them shorter (as the prop names do not get minified at all).

It's a cost of using ES6 classes (we can stop doing that at anytime), you can do some kind of hungarian notation here, but `domChangedMap` actually means nothing to what is stored in there

I'm ok with `_pendingChangedParents`

src/Utils.ts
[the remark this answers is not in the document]
ling1726: > The list is to be able to go through them (as WeakMap doesn't support enumeration)

If a regular Map is used, then you can iterate through keys there.

> I just think it's a good idea to make sure things are explicitly clear on dispose.

My suggestion does not change this

> last thing you want is to be constantly chased by the people investigating memory leaks and blaming you because you hold a strong reference

From what I see here is that the contract to remove dummy inputs `remove`, if something else happens to remove the reference that's actually bug, not  a feature.

ling1726 (approved):

src/Utils.ts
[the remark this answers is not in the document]
ling1726: yeah it's a bit of an overhead I agree, but then why are we trying to encapsulate APIs and constructs when we just go and end up passing raw HTML elements anyway?

I agree about the performance, but until we have more data there it's impossible to make either a claim for or against performance maybe we should add perf tests as a task

src/Utils.ts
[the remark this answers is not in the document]
ling1726: Yeah that makes sense to me. We control these HTMLElements, if there is a memory leak that's an indicator that _**we have a bug**_

src/Utils.ts
```diff
@@ -986,12 +986,103 @@ function setDummyInputDebugValue(
...
+        const cb = dummyInputElements.get(dummy);
+
+        if (cb) {
+            dummyInputElements.delete(dummy);
+
+            if (dummyInputElements.size === 0) {
+                delete this.domChanged;
+            }
+        }
+    }
```
ling1726, line 1018:
```suggestion
        if (dummyInputElements.has(dummy)) {
            dummyInputElements.delete(dummy);

            if (dummyInputElements.size === 0) {
                delete this.domChanged;
            }
        }
    }
```

PR #256, opened 2023-03-23
fix(Mover, Groupper): Properly prioritizing grouppers and movers in the context when they are on the same element.

mshoho:
When Mover and Groupper are set on the same element, they should be prioritized properly:

If the context element equals to the element which has both Mover and Groupper, Mover should be ignored. Otherwise Mover shouldn't be ignored and `isGroupperFirst` should be false (if there is no deeper Groupper in the context).

### DIFF
--- a/src/Root.ts
+++ b/src/Root.ts
@@ -486,9 +486,14 @@ export class RootAPI implements Types.RootAPI {
                 }
             }
 
-            if (!mover && curMover && (!modalizer || curModalizer)) {
+            if (
+                !mover &&
+                curMover &&
+                (!modalizer || curModalizer) &&
+                (!curGroupper || curElement !== element)
+            ) {
                 mover = curMover;
-                isGroupperFirst = !!groupper;
+                isGroupperFirst = !!groupper && groupper !== curGroupper;
             }
 
             if (tabsterOnElement.root) {
--- a/tests/MoverGroupper.test.tsx
+++ b/tests/MoverGroupper.test.tsx
@@ -1382,4 +1382,75 @@ describe("MoverGroupper", () => {
                 expect(el?.textContent).toEqual("Button1Button2Button3Button4");
             });
     });
+
+    it("should properly handle groupper and mover on the same element", async () => {
+        await new BroTest.BroTest(
+            (
+                <div {...getTabsterAttribute({ root: {} })}>
+                    <div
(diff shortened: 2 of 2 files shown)


ling1726 (approved):
I don't understand the changes but I understand the test 🤷

mshoho:
> I don't understand the changes but I understand the test 🤷

The PR description literally tells about the change in more words than the actual change is in the code. What it is you don't understand. I can describe more.

bsunderhus:
> > I don't understand the changes but I understand the test 🤷
> 
> The PR description literally tells about the change in more words than the actual change is in the code. What it is you don't understand. I can describe more.

Readability is damaged. It is hard to understand what the changes are by reading the code, the code is not telling what it does by itself.

src/Root.ts
```diff
@@ -486,9 +486,14 @@ export class RootAPI implements Types.RootAPI {
...
-            if (!mover && curMover && (!modalizer || curModalizer)) {
+            if (
+                !mover &&
+                curMover &&
+                (!modalizer || curModalizer) &&
+                (!curGroupper || curElement !== element)
+            ) {
                 mover = curMover;
-                isGroupperFirst = !!groupper;
+                isGroupperFirst = !!groupper && groupper !== curGroupper;
```
bsunderhus, line 496:
```suggestion
                isGroupperFirst = groupper !== curGroupper;
```
Wouldn't this be enough?

bsunderhus:
> I don't understand the changes but I understand the test 🤷

Agree, I got the sensation some variables are not named in a meaningful way to carry with them what they're verifying/doing?!

src/Root.ts
mshoho: No, because `groupper` can be undefined.

mshoho:
Let's find out. This function computes the context.
You give it an element and it finds out if the element is inside Mover, Groupper and things like that.

Example 1:
```
<div mover>
  <div groupper>
    <button>Button</button>
  </div>
</div>
```
Example 2:
```
<div groupper>
  <div mover>
    <button>Button</button>
  </div>
</div>
```

To handle the various nested scenarios properly you need to know what is closer the button. `isGroupperFirst` flag will be the difference in the button's context for the example 1 and example 2. That is the meaning of that particular flag.

And after that there corner cases begin.

PR #257, opened 2023-03-23
fix(mover): Adjacent controlled movers fix

mshoho:
The check for directly nested movers wasn't considering adjacent movers.


src/Focusable.ts
```diff
@@ -420,10 +420,9 @@ export class FocusableAPI implements Types.FocusableAPI {
             if (
                 moverElement &&
                 fromMoverElement &&
+                fromMoverElement.contains(moverElement) &&
```
ling1726, line 423: so this is for adjacent movers that are nested in another mover?

PR #262, opened 2023-04-04
fix(Core): Maintaining the initialization callbacks queue to be able to drain it on demand.

mshoho:
Normally, the initialization of tabster happens in two steps: 
1. Create all necessary API object instances;
2. Init them on the next tick when all requested APIs are created.

That didn't work in a scenario when we create the tabster core, then the required API objects, then start using the APIs in the same tick.

This PR creates a queue of all necessary initialization callbacks. And drains it from the getTabsterContext to make sure everything created is initialized in the same tick scenario.


ling1726 (approved):

ling1726:
FYI @bsunderhus approving since you tested this on FUI Dialog and resolved the issue

PR #264, opened 2023-04-04
feat(DummyInputs): Optimizing the dummy transform offsets to reduce the reflows.

mshoho:
`getComputedStyle()` causes reflow, accessing `element.scrollTop` and `element.scrollLeft` causes reflow. Setting `element.style.top` and `element.style.left` in between the above makes it all even worse.

This PR:
1. Caches the result of the things that cause reflow for the bulk during the update.
2. Splits computing the dummy input position and setting it into two phases to do the bulk computation first and the bulk position update next.
3. Improves the data structures usage in the computation.


ling1726:
How do you know these changes improve perf? have you taken any profiles or before vs after?

ling1726 (approved):

PR #268, opened 2023-04-12
feat(Mover): Making hasDefault true by default.

mshoho:
As previously concluded, it is better to have Mover's hasDefault property set to true by default.


bsunderhus (approved):

mshoho:
Setting Mover's hasDefault property set to true by default.

PR #271, opened 2023-04-19
feat(Mover): nextFor property.

mshoho:
Mover now has `nextFor` property. A selector. If an arrow key is pressed outside of the Mover and the currently focused element matches the `nextFor` selector, the Mover will be focused.


src/Mover.ts
```diff
@@ -796,24 +834,103 @@ export class MoverAPI implements Types.MoverAPI {
...
+            // the arrow keys are pressed on an element that matches the
+            // nextFor selector. Here we find if there is a Mover that wants
+            // to gain focus given the currently focused element.
+            const moversFor = this._moversFor;
+            const allMoversFor: [HTMLElement | undefined, Mover?][] = [];
+            const focusedOrBody = focused || win.document.body;
+
+            for (const nextFor of Object.keys(moversFor)) {
+                if (matchesSelector(focusedOrBody, nextFor)) {
+                    for (const m of moversFor[nextFor]) {
```
bsunderhus, line 860: `m`? as in `mover`?...

src/Mover.ts
```diff
@@ -796,24 +834,103 @@ export class MoverAPI implements Types.MoverAPI {
...
 
-        if (
-            !ctx ||
-            !ctx.mover ||
-            ctx.isExcludedFromMover ||
-            ctx.ignoreKeydown(event)
-        ) {
+        const mover = ctx?.mover;
+
+        if (!mover) {
```
bsunderhus, line 849: 4 level deep nesting inside this statement. You got me lost here 😵‍💫, can we break this into more functions or reduce the nesting somehow with some short circuiting?

src/Mover.ts
```diff
@@ -796,24 +834,103 @@ export class MoverAPI implements Types.MoverAPI {
...
+                if (focusedIndex === 0) {
+                    moverToFocus = allMoversFor[1][1];
+                } else if (focusedIndex === allMoversFor.length - 1) {
+                    moverToFocus = allMoversFor[focusedIndex - 1][1];
+                } else {
+                    moverToFocus =
+                        allMoversFor[
+                            isBackward ? focusedIndex - 1 : focusedIndex + 1
+                        ][1];
+                }
```
bsunderhus, line 904: Perhaps this can be broken into a `initializeMoverTofocus` helper method?! To help reducing nesting

src/Mover.ts
```diff
@@ -796,24 +834,103 @@ export class MoverAPI implements Types.MoverAPI {
...
+                        );
+
+                        FocusedElementState.isTabbing = true;
+                        const next = moverToFocus.findNextTabbable(
+                            fromElement,
+                            isBackward
+                        );
+                        FocusedElementState.isTabbing = false;
+
+                        next?.element?.focus();
```
bsunderhus, line 922: Perhaps this can be broken into a `focusNextElementFromMover` helper method? To help reducing nesting

src/Types.ts
```diff
@@ -616,6 +617,14 @@ export interface MoverProps {
      * @default 0.8
      */
     visibilityTolerance?: number;
+    /**
+     * CSS selector. If the currently focused element matches this selector,
```
bsunderhus, line 621:
```suggestion
     * A [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors). If the currently focused element matches this selector,
```

src/Mover.ts
```diff
@@ -763,6 +799,90 @@ export class MoverAPI implements Types.MoverAPI {
         }
     };
 
+    private _focusMoverFor(
```
ling1726, line 802: Why is this feature necessary?
 
- I want arrow navigation inside a DOM element - I apply mover
- I press an arrow key outside the mover
- focus should go into the mover

Why does this become a platform problem?

src/Mover.ts
```diff
@@ -763,6 +799,90 @@ export class MoverAPI implements Types.MoverAPI {
...
+                    const element = mover.getElement();
+
+                    if (element) {
+                        allMoversFor.push({ element, mover });
+                    }
+                }
+            }
+        }
+
+        if (allMoversFor.length > 0) {
```
ling1726, line 828: you don't need the extra sort right? you could just check for the closest mover while you are matching selectors.

src/Mover.ts
mshoho: We're not perfect yet and allow focus go to <body> from time to time. The current version of Teams has this helper that if you're seeing a chat, and you press arrow keys, your focus will be moved from <body> to the chat regiving you a meaningful starting point (otherwise you would need to Tab through the entire app again and that's not the best experience). So, it will be possible to just set `nextFor: 'body'` on the chat's mover and reuse Tabster's machinery to handle that case.

I don't think it's a part of some standard but as far as I'm aware of, there are multiple people (including the screen reader users) who find this behaviour very helpful. It is easy to implement on the Tabster level and it is guaranteed to be consistently compatible with all our focus management primitives, giving more stability comparing to the application level implementation.

Another possible use case is you might be inside an input, but you might want to move focus to some choices selection list below or above. That scenario will also be solvable with one Tabster property.

PR #286, opened 2023-06-15
feat(Modalizer): A createTabster() option to provide alwaysAccessibleSelector selector for the elements to not set aria-hidden on.

mshoho:
By default everything outside of the active Modalizer gets aria-hidden attribute. We are adding option to be able to exclude some particular elements (for example aria-live announcement containers) from receiving aria-hidden.

### DIFF
--- a/src/Modalizer.ts
+++ b/src/Modalizer.ts
@@ -329,18 +329,20 @@ export class ModalizerAPI implements Types.ModalizerAPI {
     private _augMap: WeakMap<HTMLElement, true>;
     private _aug: WeakRef<HTMLElement>[];
     private _hiddenUpdateTimer: number | undefined;
+    private _alwaysAccessibleSelector: string | undefined;
 
     activeId: string | undefined;
     currentIsOthersAccessible: boolean | undefined;
     activeElements: WeakRef<HTMLElement>[];
 
-    constructor(tabster: Types.TabsterCore) {
+    constructor(tabster: Types.TabsterCore, alwaysAccessibleSelector?: string) {
         this._tabster = tabster;
         this._win = tabster.getWindow;
         this._modalizers = {};
         this._parts = {};
         this._augMap = new WeakMap();
         this._aug = [];
+        this._alwaysAccessibleSelector = alwaysAccessibleSelector;
         this.activeElements = [];
 
         if (!tabster.controlTab) {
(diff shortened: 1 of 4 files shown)


src/Types.ts
```diff
@@ -36,6 +36,13 @@ export interface TabsterCoreProps {
      * This option allows to enable dummy inputs on Root.
      */
     rootDummyInputs?: boolean;
+    /**
+     * When Modalizer is active, we put aria-hidden to everything else to hide
+     * it from screen readers. This CSS selector allows to exclude some elements
+     * from this behaviour. For example, this could be used to exclude aria-live region
+     * with the application-wide status announcements.
+     */
+    modalizerAlwaysAccessible?: string;
```
ling1726, line 45: why is this an overall tabster property and not a specific modalizer property?

mshoho: because it is for the global application scenario. and having it here wouldn't require refactoring every modalizer used. and it basically is a workaround for the lack of browser functions (no JavaScript API to announce things via screen reader), so it will be good to be able to revoke it easily once the bright future. in short, the Modalizer is a legit application logic scenario. and aria-live containers in the body is just a workaround that is hoped to be temporarty.

ling1726: but users have to initialize the modalizer with `getModalizer` right? why not add the configuration there?

src/Modalizer.ts
```diff
@@ -654,7 +656,13 @@ export class ModalizerAPI implements Types.ModalizerAPI {
         const parts = this._parts;
         const visibleElements: HTMLElement[] = [];
         const hiddenElements: HTMLElement[] = [];
-        const alwaysAccessibleElements: HTMLElement[] = [];
+        const alwaysAccessibleSelector = this._alwaysAccessibleSelector;
+        const alwaysAccessibleElements: HTMLElement[] = alwaysAccessibleSelector
```
ling1726, line 660: Is it safe to pull this on modalizer creation only? what if there are more always accessible elements added after the modalizer is created?

src/Modalizer.ts
```diff
@@ -654,7 +656,13 @@ export class ModalizerAPI implements Types.ModalizerAPI {
...
+            ? Array.prototype.slice.call(
+                  body.querySelectorAll(alwaysAccessibleSelector),
+                  0
+              )
```
ling1726, line 664:
```suggestion
            ? Array.from(
                  body.querySelectorAll(alwaysAccessibleSelector))
```

ling1726 (approved):

PR #287, opened 2023-06-19
Use getAttribute instead of accessing id directly

lgoeckener:
If there is an input with name "id" in a form, element.id will target that input field instead of the id attribute of the form.

```html
<form id="foo">
    <input type="hidden" name="id" value="1" />
</form>
```

Means, if in the current iteration of buildElementSelector, the `element` is the `<form>`-element,

```js
element.id
```

will point to the input.

Therefore using `element.getAttribute("id")` will make sure, that the id-attribute of the current element is used.

### DIFF
--- a/src/Deloser.ts
+++ b/src/Deloser.ts
@@ -296,9 +296,11 @@ function buildElementSelector(
     const escapeRegExp = /(:|\.|\[|\]|,|=|@)/g;
     const escapeReplaceValue = "\\$1";
 
-    if (element.id) {
+    const elementId = element.getAttribute("id");
+
+    if (elementId) {
         selector.push(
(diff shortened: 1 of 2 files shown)


mshoho:
@lgoeckener thanks! Could you please add a test for the case?

lgoeckener:
@mshoho I would love to! But I'm afraid I'm not quite sure in how to implement the test in the given environment.

I'm not familiar with the BroTest testing environment nor how I could create a scenario that would end up testing that particular branch.
Besides that I cannot create DOM elements since packages like JSDOM or react-testing-library are not installed.

If you give permission to install additional packages just for testing this branch (I would also need to export the buildElementSelector function), I could proceed doing so in a way that I'm familiar with.

lgoeckener:
@microsoft-github-policy-service agree

PR #294, opened 2023-07-14
feat: Implement Restorer API

ling1726:
Adds the new Restorer API which is a specialized deloser that only restores focus to a specific target element when focus has been lost from a specific source element.

### DIFF
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -87,6 +87,7 @@ export function updateTabsterByAttribute(
                 case "root":
                 case "groupper":
                 case "modalizer":
+                case "restorer":
                 case "mover":
                     // eslint-disable-next-line no-case-declarations
                     const part = tabsterOnElement[key];
@@ -175,6 +176,17 @@ export function updateTabsterByAttribute(
                         );
                     }
                 }
+                break;
+
+            case "restorer":
+                // TODO implement setProps on restorer API
+                if (newTabsterProps.restorer && tabster.restorer) {
+                    tabsterOnElement.restorer = tabster.restorer.createRestorer(
+                        element,
+                        newTabsterProps.restorer
+                    );
+                }
+
                 break;
 
             case "focusable":
--- a/src/Restorer.ts
+++ b/src/Restorer.ts
@@ -1,0 +1,130 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import type {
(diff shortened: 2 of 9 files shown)


src/Instance.ts
```diff
@@ -177,6 +178,17 @@ export function updateTabsterByAttribute(
                 }
                 break;
 
+            case "restorer":
```
mshoho, line 181: Seems like the dispose part is missing (it's in the same file).  And for the sake of consistency, let's do `if (__DEV__)` with the error when Restorer is not initialized yet.

mshoho: Oops. I have missed line 90.

src/Restorer.ts
```diff
@@ -0,0 +1,143 @@
...
+            this._restoreFocus(e.target as HTMLElement)
+        );
+    };
+
+    private _onFocusIn = (element: HTMLElement | undefined) => {
+        if (!element) {
+            return;
+        }
+
+        const tabsterAttribute = getTabsterAttributeOnElement(element);
```
mshoho, line 98: There is getTabsterOnElement which is used all over and that takes the already created entities from storage instead of parsing them every time. I makes sense to add getProps() to Restorer (like all other things have) and instead of JSON.parse on every focusin it will be just picking an already parsed object from the storage.

ling1726: Updated

mshoho (approved):

src/Restorer.ts
```diff
@@ -0,0 +1,146 @@
...
+    FocusedElementState,
+    TabsterCore,
+} from "./Types";
+import { RestorerTypes } from "./Types";
+import { TabsterPart } from "./Utils";
+
+const EVENT_NAME = "restorer:restorefocus";
+const HISOTRY_DEPTH = 10;
+
+class Restorer extends TabsterPart<RestorerProps> implements RestorerInterace {
```
mshoho, line 22: A little typo in `RestorerInterace`.

src/Restorer.ts
```diff
@@ -0,0 +1,146 @@
...
+            !this._keyboardNavState.isNavigatingWithKeyboard() &&
+            // Source no longer exists on DOM - always restore focus
+            doc.body.contains(Source)
+        ) {
+            return;
+        }
+
+        let weakRef = this._history.pop();
+        while (!doc.body.contains(weakRef?.deref()?.parentElement ?? null)) {
+            console.log("loop");
```
mshoho, line 136: Oops.

PR #295, opened 2023-07-17
feat(Modalizer): An accessibleCheck callback for getModalizer() to be able to avoid setting aria-hidden on specific elements.

mshoho:
Active Modalizer sets `aria-hidden` to everything outside to hide it from screen readers. We already have `alwaysAccessibleSelector` selector to avoid setting `aria-hidden` on some elements that match a specific selector. This is another option to do the same but programmatically.

### DIFF
--- a/src/Modalizer.ts
+++ b/src/Modalizer.ts
@@ -330,12 +330,17 @@ export class ModalizerAPI implements Types.ModalizerAPI {
     private _aug: WeakRef<HTMLElement>[];
     private _hiddenUpdateTimer: number | undefined;
     private _alwaysAccessibleSelector: string | undefined;
+    private _accessibleCheck: Types.ModalizerElementAccessibleCheck | undefined;
 
     activeId: string | undefined;
     currentIsOthersAccessible: boolean | undefined;
     activeElements: WeakRef<HTMLElement>[];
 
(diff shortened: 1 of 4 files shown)


src/Modalizer.ts
```diff
@@ -330,19 +330,25 @@ export class ModalizerAPI implements Types.ModalizerAPI {
...
+    private _accessibleCheck: Types.ModalizerElementAccessibleCheck | undefined;
 
     activeId: string | undefined;
     currentIsOthersAccessible: boolean | undefined;
     activeElements: WeakRef<HTMLElement>[];
 
-    constructor(tabster: Types.TabsterCore, alwaysAccessibleSelector?: string) {
+    constructor(
+        tabster: Types.TabsterCore,
+        alwaysAccessibleSelector?: string,
```
ling1726, line 341: Let's add @deprecate on `alwaysAccessibleSelector`, there's no reason to have it with the check API

mshoho: I can just remove it and release next major version.

ling1726: I don't think it's a big enough change to warrant a major version, but I can leave that up to you

ling1726: So I'd prefer just to deprecate this, until we actually have a feature that's worth the cost of a major bump

ling1726 (approved):

PR #301, opened 2023-08-15
fix: use FakeWeakRef when WeakRef is unavailable

cjsheets:
FakeWeakRef is a useful construct for clients that lack WeakRef support. In particular, there are still many Android devices that don't have WeakRef available. 

This change adds FakeWeakRef fallback to Restorer and Modalizer.

### DIFF
--- a/src/Modalizer.ts
+++ b/src/Modalizer.ts
@@ -18,6 +18,10 @@ import {
     WeakHTMLElement,
     triggerEvent,
     augmentAttribute,
+    getInstanceContext,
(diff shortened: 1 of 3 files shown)


cjsheets:
I see, that makes sense. We support the last 4 major Android versions and 2 major iOS versions.

From what we've seen WeakRef support is present in Android browser versions >=84, consistent with the support matrix you linked to. Since 4 versions ago was Android 11 and Chrome 84 was released before Android 11, it sounds like our supported devices are in alignment.

PR #302, opened 2023-08-20
feat: Out of DOM order focusing improvement

mshoho:
Up to this moment we were trying to find out if the next focusable element is out of DOM order (like when are in a modalizer) after the next element search. This detection is needed to handle corner cases in uncontrolled mode when we need to use default action to move focus outside of the browser window, skipping the out of scope focusables. This after the fact detection was hard to understand and was not covering all corner cases. To improve and simplify the overall handling, we now detect if the element is out of DOM order right during the search. So that we can move focus programmatically when next focusable is out of DOM order. This change will also allow to remove the majority of dummy inputs in the subsequent pull requests.


CHANGELOG.md
```diff
@@ -4,9 +4,17 @@ All notable changes to this project will be documented in this file. Dates are d
 
 Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
+#### [v4.8.0-canary.0](https://github.com/microsoft/tabster/compare/v4.7.2...v4.8.0-canary.0)
+
+-   feat: Improving out of DOM order next focusables handling. [`d390d1d`](https://github.com/microsoft/tabster/commit/d390d1d777854db3e1aeac24c0ff1016519caeab)
+-   Oops. [`1a86249`](https://github.com/microsoft/tabster/commit/1a86249b50e157493ff20491fd90cf7d7026ee6a)
```
bsunderhus, line 10: Oops?

src/Focusable.ts
```diff
@@ -139,16 +139,17 @@ export class FocusableAPI implements Types.FocusableAPI {
...
 
     findPrev(options: Types.FindNextProps): HTMLElement | null | undefined {
-        return this.findElement({
-            isBackward: true,
-            ...options,
-        });
+        const opts = { ...options, isBackward: true };
+        const ret = this.findElement(opts);
+        options.outOfDOMOrderResult = opts.outOfDOMOrderResult;
+        return ret;
```
bsunderhus, line 152: Not sure I'm following what's happening here, seems like `findElement` is doing way more than just finding elements?!

src/Groupper.ts
```diff
@@ -166,35 +169,46 @@ export class Groupper
...
+                ? tabster.focusable.findPrev(
+                      (findProps = {
+                          container: groupperElement,
+                          currentElement,
+                          referenceElement,
+                          onUncontrolled,
+                          ignoreUncontrolled,
+                          ignoreAccessibility,
+                          useActiveModalizer: true,
+                      })
```
bsunderhus, line 197: Can we re-write this assignments to something more readable?! usually using assignments as expressions is a bit convoluted.

Something like:
```ts
if (this._shouldTabInside && groupperFirstFocusable) {
  const findProps: Types.FindNextProps = {
    container: groupperElement,
    currentElement,
    referenceElement,
    onUncontrolled,
    ignoreUncontrolled,
    ignoreAccessibility,
    useActiveModalizer: true,
  };

  if (isBackward) {
    next = tabster.focusable.findPrev(findProps);
  } else {
    next = tabster.focusable.findNext(findProps);
  }

  outOfDOMOrder = !!findProps.outOfDOMOrderResult;

  if (!uncontrolled && !next && !outOfDOMOrder) {
    next = tabster.focusable.findNext({
      container: groupperElement,
      currentElement: null,
      referenceElement,
      onUncontrolled,
      ignoreUncontrolled,
      ignoreAccessibility,
      useActiveModalizer: true,
    });

    outOfDOMOrder = true;
  }
}
```

src/Focusable.ts
mshoho: Why? It finds elements and returns metadata about the found elements.

CHANGELOG.md
mshoho: That's not coming to the final merge. We undo the canary release commits and the squashed merge to master will only have the PR description.

src/Focusable.ts
```diff
@@ -139,16 +139,17 @@ export class FocusableAPI implements Types.FocusableAPI {
     }
 
     findNext(options: Types.FindNextProps): HTMLElement | null | undefined {
-        return this.findElement({
-            ...options,
-        });
+        const opts = { ...options };
+        const ret = this.findElement(opts);
+        options.outOfDOMOrderResult = opts.outOfDOMOrderResult;
```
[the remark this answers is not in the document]
ling1726: You will also be able to rename from `outOfDOMOrderResult` to `outOfDOMORder`

ling1726 (approved):

PR #304, opened 2023-08-31
Revaming handling for uncontrolled areas.

mshoho:
This adjusts how the uncontrolled areas are handled. We have the ability to set `data-tabster={uncontrolled: {}}` on an element and Tabster won't handle tabbing inside those areas. We were treating those areas as black boxes before. And we had workarounds for corner cases. Like `ignoreUncontrolled` prop for finding focusables and duplicate search (with this prop disabled and then enabled). This PR changes the approach to handling the uncontrolled areas. Instead of treating uncontrolled areas as black boxes with special handling in multiple places, we just have the flag in the element context and return that flag with `find*` functions and it is handled by `findNextTabbable` functions when relevant.

To make this approach fully workable, we also now have a concept of phantom dummy with fixed target. When Tab is pressed in an uncontrolled area, and there is something tabster-related inside that area (like a mover inside uncontrolled) which would like to move focus to something out of DOM order, we add a phantom dummy input to redirect focus to that area. If something that controls uncontrolled area decides to handle Tab, this dummy won't be focused, but if the default action kicks in, it will be focused and will move focus to a proper next element. And in any case the phantom dummy will disappear with the next tick.

And lastly, as the context has been adjusted, a few other props in the context have been renamed to be hopefully more understandable (and to not have `is` prefix which some props have and some don't, making it more consistent).


src/Focusable.ts
```diff
@@ -202,12 +218,15 @@ export class FocusableAPI implements Types.FocusableAPI {
         const elements: HTMLElement[] = [];
 
         let { acceptCondition } = options;
+        let hasCustomCondition = true;
```
ling1726, line 221:
```suggestion
        const hasCustomCondition = !!acceptCondition;
```

This can be constant, there's no reason for this to be assigned later

src/Focusable.ts
```diff
@@ -248,7 +267,9 @@ export class FocusableAPI implements Types.FocusableAPI {
         const prepareForNextElement = (
             shouldContinueIfNotFound?: boolean
         ): boolean => {
-            const foundElement = acceptElementState.foundElement;
+            const foundElement =
+                acceptElementState.foundElement ||
+                acceptElementState.foundBackward;
```
ling1726, line 272:
```suggestion
            const foundElement =
                acceptElementState.foundElement ??
                acceptElementState.foundBackward;
```

src/Focusable.ts
```diff
@@ -378,35 +395,32 @@ export class FocusableAPI implements Types.FocusableAPI {
             return NodeFilter.FILTER_SKIP;
         }
 
-        if (state.ignoreUncontrolled) {
-            if (shouldIgnoreFocus(element)) {
-                return NodeFilter.FILTER_SKIP;
+        if (shouldIgnoreFocus(element)) {
+            if (
+                this._tabster.focusable.isFocusable(
```
ling1726, line 400:
```suggestion
                this.isFocusable(
```

This is a member of the same class, why do you need to accecss it through a tabster proxy?

CHANGELOG.md
```diff
@@ -4,9 +4,27 @@ All notable changes to this project will be documented in this file. Dates are d
 
 Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
+#### [v5.0.0-canary.0](https://github.com/microsoft/tabster/compare/v4.8.0-canary.0...v5.0.0-canary.0)
```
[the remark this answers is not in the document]
mshoho: I do canary releases from this branch for testing. I will revert the canary commits once everything is working fully properly.

src/State/FocusedElement.ts
```diff
@@ -540,65 +482,31 @@ export class FocusedElementState
...
-                    DummyInputManager.moveWithPhantomDummy(
-                        this._tabster,
-                        uncontrolled,
-                        moveOutside,
-                        isBackward
-                    );
-                }
+        if (nextElement) {
+            if (ctx.uncontrolled && controlTab) {
+                // We are in uncontrolled zone. We allow whatever controls it to move
```
ling1726, line 492: Have you tested this behaviour with older FUI focus utilities like FocusZone and FocusTrapZone?

src/Utils.ts
```diff
@@ -957,6 +973,33 @@ export class DummyInputManager {
             }
         }
     }
+
+    static addPhantomDummyWithTarget(
+        tabster: Types.TabsterCore,
+        anchor: HTMLElement,
+        isBackward: boolean,
+        element: HTMLElement
```
ling1726, line 981: what is `element` in this context?

src/State/FocusedElement.ts
mshoho: I am currently fixing a bug exactly in this place. It was caught by TMP automation in the FocusZone scenarios. And I will of course run the FUI automation to make sure it is green. Will also add more tests to cover the caught corner cases.

src/Utils.ts
mshoho: I can rename it to targetElement.

ling1726: can you describe the relationship between `anchor` and `targetElement`? I still can't quite understand it

mshoho: Anchor is the element we're moving from. Practically, the currently focused element. The phantom dummy will be inserted before or after this element (depending on the direction of travel and I am fixing a bug that for the cases with nested focusables it sometimes should be inserted inside the anchor element. Basically, it should be physically the element the browser will pick to focus if default action for Tab press is allowed. If something that controls the uncontrolled area picks an element to focus, so be it, the phantom dummy will just disappear with the next tick. But if it allows the default action, our dummy will receive focus and transfer it to the target element (and of course will also disappear with the next tick).

mshoho: Renamed it to sourceElement and targetElement.

src/Root.ts
```diff
@@ -203,54 +200,42 @@ export class Root
...
+    private _onFocusIn = (event: FocusEvent) => {
+        const target = event.target as HTMLElement | null;
 
-            if (!ctx || ctx.uncontrolled || this._tabster.rootDummyInputs) {
-                this._dummyManager?.setTabbable(false);
-                return;
-            }
-        } else {
-            this._setFocused(false);
+        if (target && this._element.get()?.contains(target)) {
```
ling1726, line 232: It's possible that the root is a virtual parent of the focused element and not a direct DOM parent. Do you know if that that would cause problems if it doesn't uses the changes in https://github.com/microsoft/tabster/pull/308?

src/Uncontrolled.ts
```diff
@@ -10,7 +10,13 @@ import * as Types from "./Types";
...
 export class UncontrolledAPI implements Types.UncontrolledAPI {
-    constructor() {
-        /**/
+    private _isTrappingFocus?: (element: HTMLElement) => boolean;
+
+    constructor(isTrappingFocus?: (element: HTMLElement) => boolean) {
+        this._isTrappingFocus = isTrappingFocus;
+    }
+
+    isTrappingFocus(element: HTMLElement): boolean {
```
ling1726, line 19: Is this supposed to be used with FocusTrapZone?

ling1726 (approved):

src/Utils.ts
```diff
@@ -919,42 +932,113 @@ export class DummyInputManager {
         const input = dummy.input;
 
         if (input) {
-            const parent = element.parentElement;
+            let parent: HTMLElement | null;
```
ling1726, line 935: Please try and rethink the code here, I can't understand what's going on here after multiple reads

mshoho: I have added a comment describing the phantom dummy concept.

src/Root.ts
mshoho: Added getParent(). And I want to go through the virtual parent scenarios deeper next. To create more tests as I suspect we might have corner cases.

PR #306, opened 2023-09-15
fix: Restorer should restore focus when element is removed in Safari & Firefox

### LINKED ISSUE #305: Restorer relies on focusout event firing when element is removed
Safari does not fire `focusout` or any other events when a node is removed, which means focus isn't restored when the restorer source is unmounted, even though focus returns to the body.

This can be reproduced in the Restorer stories in Safari, and also in FUI in components with popups (e.g. Popover, Dialog, Menu, etc.).

smhigley:
Fixes #305

Not sure if this is the best approach, but it does fix the root issue. Some of the constraints are:

- `document.activeElement` has already been set to `body` by the time the MutationObserver removedNodes mutation is detected
- Non-Chromium browsers fire literally no events when the focused element is removed from the DOM & `activeElement` is set to `body`
- This could in theory be fixed by firing `.blur()` on the `document.activeElement` _before_ it's removed, if we can figure out any way to do so

Here's a reduced demo of the basic issue in Safari & Firefox: https://jsfiddle.net/5pcf9L64/

### DIFF
--- a/src/Restorer.ts
+++ b/src/Restorer.ts
@@ -29,21 +29,19 @@ class Restorer extends TabsterPart<RestorerProps> implements RestorerInterface {
     ) {
         super(tabster, element, props);
 
-        // need to attach MutationObserver for Safari to detect source element removal.
-        // removing the activeElement does not trigger a focusout event in Safari
-        const isSafari = navigator?.userAgent.indexOf('Safari') !== -1 && navigator?.userAgent.indexOf('Chrome') === -1;
-
         if (this._props.type === RestorerTypes.Source) {
             const element = this._element?.get();
             element?.addEventListener("focusout", this._onFocusOut);
 
-            // attach MutationObserver to parent for Safari
+            // Need to also listen to node removals, which do not trigger focus events in non-Chromium browsers
             const parent = element?.parentElement;
-            if (isSafari && parent) {
+            if (parent) {
                 this._observer = new MutationObserver(this._onNodeRemoved);
-                this._observer.observe(parent, { childList: true, subtree: false });
+                this._observer.observe(parent, {
+                    childList: true,
+                    subtree: false,
+                });
             }
-
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/Restorer.ts
```diff
@@ -20,15 +20,28 @@ const EVENT_NAME = "restorer:restorefocus";
 const HISOTRY_DEPTH = 10;
 
 class Restorer extends TabsterPart<RestorerProps> implements RestorerInterface {
+    private _observer: MutationObserver | undefined;
```
ling1726, line 23: This would start a mutation observer per instance of Restorer - this seems kind of scary performance wise.

We could only have one mutation observer instance on RestorerAPI, to share one mutation observer instance among all Restorer instances

src/Restorer.ts
```diff
@@ -20,15 +20,28 @@ const EVENT_NAME = "restorer:restorefocus";
...
             const element = this._element?.get();
             element?.addEventListener("focusout", this._onFocusOut);
+
+            // Need to also listen to node removals, which do not trigger focus events in non-Chromium browsers
+            const parent = element?.parentElement;
+            if (parent) {
+                this._observer = new MutationObserver(this._onNodeRemoved);
+                this._observer.observe(parent, {
+                    childList: true,
+                    subtree: false,
```
ling1726, line 42: why is subtree false? a focusable node removal could happen anywhere in the tree

ling1726: In this case, there is no need to have a mutation observer per Restorer instance. Tabster already has a mutation observer that will call `dispose` elements with tabster on them. You just need to update the `dispose` function like so 

```ts
dispose(): void {
        if (this._props.type === RestorerTypes.Source) {
            const doc = this._tabster.getWindow().document;
            const element = this._element?.get();
            element?.removeEventListener("focusout", this._onFocusOut);

            if (element && !doc.body.contains(element)) {
                doc.dispatchEvent(
                    new Event(EVENT_NAME, {
                        bubbles: true,
                    })
                );
            }
        }
    }
```

This should fix the issue for all browsers when the element with Restorer it is removed from DOM.

For elements inside the Restorer - the RestorerAPI could have a mutation observer all the Restorer instances (this should probably be opt-in for Restorer through configuration).

src/Restorer.ts
```diff
@@ -49,6 +62,26 @@ class Restorer extends TabsterPart<RestorerProps> implements RestorerInterface {
...
+
+    private _onNodeRemoved = (mutations: MutationRecord[]) => {
+        for (const mutation of mutations) {
+            for (let i = 0; i < mutation.removedNodes.length; i++) {
+                const element = this._element?.get();
+                if (mutation.removedNodes[i] === element) {
+                    const lastFocused =
+                        this._tabster.focusedElement.getFocusedElement();
+                    if (lastFocused && element.contains(lastFocused)) {
+                        document.body?.dispatchEvent(
```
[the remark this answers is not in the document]
smhigley: good catch, forgot to update that 😅

PR #308, opened 2023-09-25
feat: getParent prop for createTabster

ling1726:
Adds a new `getParent` prop for `createTabster. This lets users override how to get a parent of a HTMLElement, which is useful for libraries like Fluent UI where the parentElement does not need to follow the DOM order.

### DIFF
--- a/src/Tabster.ts
+++ b/src/Tabster.ts
@@ -142,7 +142,7 @@ class TabsterCore implements Types.TabsterCore {
      * props can/should be mergeable, so let's add more as we move on.
      * @param props Tabster props
      */
-    mergeProps(props?: Types.TabsterCoreProps) {
+    private _mergeProps(props?: Types.TabsterCoreProps) {
         if (!props) {
             return;
         }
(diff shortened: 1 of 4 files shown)


src/Tabster.ts
```diff
@@ -134,13 +137,31 @@ class TabsterCore implements Types.TabsterCore {
         });
     }
 
-    createTabster(noRefCount?: boolean): Types.Tabster {
+    /**
+     * Merges external props with the current props. Not all
+     * props can/should be mergeable, so let's add more as we move on.
+     * @param props Tabster props
+     */
+    mergeProps(props?: Types.TabsterCoreProps) {
```
mshoho, line 145: Does it need to be public? Seems like private method would do.

tests/Tabster.test.tsx
```diff
@@ -17,6 +17,69 @@ describe("Tabster dispose", () => {
...
+                // dispose default tabster on the test page
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore
+                return window.__tabsterInstance.getParent().id;
+            }, second)
+            .check((id) => {
+                expect(id).toEqual(second);
+            });
+    });
+
```
mshoho, line 82: I would also expect to have a test that tests getParent indirectly (like having some DOM with virtualParent prop). Basically test something this feature is created for.

mshoho (approved):

PR #327, opened 2023-11-20
feat(Uncontrolled): completely prop to indicate that uncontrolled area fully controls Tab.

mshoho:
Currently we have `checkUncontrolledTrappingFocus` callback in `createTabster()` to provide the flag that an uncontrolled area is trapping focus. Adding `completely` property to uncontrolled props on DOM element to be able to set this indicator right on the particular DOM elements. Also deprecating `checkUncontrolledTrappingFocus` and adding `checkUncontrolledCompletely` for the sake of consistency and better coexistence.

### DIFF
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -55,10 +55,14 @@ export interface TabsterCoreProps {
      * non-undefined value, the callback's value will dominate the element's
      * uncontrolled.completely value.
      */
-    checkUncontrolledTrappingFocus?: (
+    checkUncontrolledCompletely?: (
         element: HTMLElement,
         completely: boolean // A uncontrolled.completely value from the element.
     ) => boolean | undefined;
+    /**
+     * @deprecated use checkUncontrolledCompletely.
+     */
+    checkUncontrolledTrappingFocus?: (element: HTMLElement) => boolean;
     /**
      * Custom getter for parent elements. Defaults to the default .parentElement call
      * Currently only used to detect tabster contexts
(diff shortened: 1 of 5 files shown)


src/Types.ts
```diff
@@ -1000,6 +1000,15 @@ export type ModalizerElementAccessibleCheck = (
     activeModalizerElements?: HTMLElement[]
 ) => boolean;
 
+export interface UncontrolledProps {
+    // Indicated that uncontrolled area traps focus, that makes Tabster
+    // to give it full control over the Tab handling. If not set, Tabster
+    // might still handle Tab in the uncontrolled area, when, for example,
+    // there is an inactive Modalizer (that needs to be skipped) after the
+    // last focusable element of the uncontrolled area.
+    trapsFocus?: boolean;
```
ling1726, line 1009: If this is to let tabster have full control of Tab handling why call it `trapFocus`? maybe something like `complete` or smth like that to indicate it's completely uncontrolled

src/Types.ts
```diff
@@ -47,11 +47,11 @@ export interface TabsterCoreProps {
...
     /**
-     * A callback that will be called for the uncontrolled areas with `trapsFocus`
-     * when Tabster wants to know is the element is currently trapping focus and
-     * Tabster should not interfere with handling Tab.
+     * A callback that will be called for the uncontrolled areas when Tabster wants
+     * to know is the uncontrolled element wants complete control (for example it
+     * is trapping focus) and Tabster should not interfere with handling Tab.
      */
-    checkUncontrolledTrappingFocus?: (element: HTMLElement) => boolean;
+    checkUncontrolledCompletely?: (element: HTMLElement) => boolean;
```
ling1726, line 54: This is a breaking API change - If you deprecated this prop and created a new one, it can be released as a minor change

ling1726 (approved):

src/Types.ts
mshoho: I suppose it would be better to keep it like before for now. And I'll rename it in the next PR which will address the callbacks handling as it will also require a major version bump.

PR #329, opened 2023-11-29
feat(Focusable): Checking for display:none during the element search to skip hidden DOM.

mshoho:
When we have a chunk of hidden DOM, it makes sense to just skip it completely instead of walking through it.
Also, adjusting two cases of using offsetParent as they were not handling position:fixed properly.


ling1726 (changes requested):
Will need to profile this - since it gets computed style of elements during focusable search

mshoho:
> Will need to profile this - since it gets computed style of elements during focusable search

I totally agree.

mshoho:
> Will need to profile this - since it gets computed style of elements during focusable search

I have removed the check from acceptElement. The usage that is left should be equivalent to the usage before.

mshoho:
TMP performance gates don't seem unhappy with this change.

ling1726 (approved):

PR #330, opened 2023-12-08
feat(Groupper): Moving isIgnoredInput from Mover to utils and reusing it from Groupper.

mshoho:
We have `isIgnoredInput()` that is used to check if Mover needs to handle a key press or not for particular states of the currently focused elements. Reusing this routine for Groupper too and improving the handling logic.

### DIFF
--- a/src/Groupper.ts
+++ b/src/Groupper.ts
@@ -608,10 +608,11 @@ export class GroupperAPI implements Types.GroupperAPI {
             const focused = tabster.focusedElement.getFocusedElement();
 
             if (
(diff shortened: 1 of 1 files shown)


src/Groupper.ts
```diff
@@ -590,11 +590,11 @@ export class GroupperAPI implements Types.GroupperAPI {
         }
     };
 
-    handleKeyPress(
+    async handleKeyPress(
```
bsunderhus, line 593: Will this be a breaking change? what version will this PR be bumping?

PR #331, opened 2023-12-12
feat(Groupper): Handling Esc asynchronously to give the app to react on Esc press.

mshoho:
If the application moves focus when Esc is pressed, that means the keypress is handled already and Groupper should not interfere. This covers a part of the scenarios when we have a combobox or a popup inside that moves focus when the menu is closed.


bsunderhus (approved):
LGTM 👍🏻

PR #347, opened 2024-01-10
ShadowDOM support

mshoho:


src/Focusable.ts
```diff
@@ -293,7 +294,7 @@ export class FocusableAPI implements Types.FocusableAPI {
             out.outOfDOMOrder = true;
         }
 
-        if (currentElement) {
+        if (currentElement && dom.nodeContains(container, currentElement)) {
```
spmonahan, line 297: Does this address a bug that previously existed or was it added specifically for shadow roots?

src/Shadowdomize/DOMFunctions.ts
```diff
@@ -0,0 +1,226 @@
...
+}
+
+export function getParentNode(
+    node: Node | null | undefined
+): ParentNode | null {
+    if (!node) {
+        return null;
+    }
+
+    if (
```
spmonahan, line 56: We should either add it now or add a `TODO` to handle slotted nodes.

src/Utils.ts
```diff
@@ -1645,28 +1658,42 @@ class DummyInputManagerCore {
...
 
-            const firstElementChild = element.firstElementChild;
+            const firstElementChild = dom.getFirstElementChild(element);
 
-            if (firstElementChild && firstElementChild !== firstDummyInput) {
-                element.insertBefore(firstDummyInput, firstElementChild);
+            if (
+                firstElementChild &&
+                firstElementChild !== firstDummyInput &&
+                firstElementChild.parentNode
```
spmonahan, line 1690: Does this need to be updated to use DOMAPI?

src/Utils.ts
```diff
@@ -1645,28 +1658,42 @@ class DummyInputManagerCore {
...
+            ) {
+                dom.insertBefore(
+                    firstElementChild.parentNode,
```
spmonahan, line 1693: This too?

spmonahan (approved):
Mostly reviewing the mechanical changes, which look good!

src/Deloser.ts
```diff
@@ -340,17 +341,27 @@ function buildSelector(element: HTMLElement): string | undefined {
...
 
-    let el = element.parentElement;
+    let node = dom.getParentNode(element);
 
-    while (el) {
-        const isBody = el.tagName === "BODY";
-        selector.unshift(buildElementSelector(el, false, !isBody));
-
-        if (isBody) {
+    while (node) {
```
ling1726, line 346:
```suggestion
    while (node &&  node.nodeType !== Node.DOCUMENT_FRAGMENT_NODE) {
```

this should remove the need to check the nodeType inside the loop?

src/Groupper.ts
```diff
@@ -431,7 +439,15 @@ export class GroupperAPI implements Types.GroupperAPI {
...
         this._tabster.focusedElement.subscribeFirst(this._onFocus);
 
-        win.document.addEventListener("mousedown", this._onMouseDown, true);
+        const doc = win.document;
+
+        const activeElement = dom.getActiveElement(doc);
+
+        if (activeElement) {
+            this._onFocus(activeElement as HTMLElement);
+        }
```
ling1726, line 448: why was this added?

src/Root.ts
```diff
@@ -227,10 +230,12 @@ export class Root
...
     };
 
-    private _onFocusIn = (event: FocusEvent) => {
+    private _onFocusIn = () => {
         const getParent = this._tabster.getParent;
         const rootElement = this._element.get();
-        let curElement = event.target as HTMLElement | null;
+        let curElement = dom.getActiveElement(
+            this._tabster.getWindow().document
+        );
```
ling1726, line 239: you could also use the compsed path here?

src/Shadowdomize/DOMFunctions.ts
```diff
@@ -0,0 +1,226 @@
...
+export function getPreviousSibling(
+    node: Node | null | undefined
+): ChildNode | null {
+    if (!node) {
+        return null;
+    }
+
+    let sibling = node.previousSibling;
+
+    if (!sibling && node.parentElement?.shadowRoot) {
```
ling1726, line 128: I wonder if we need to handle this case? Nothing you append to the host of a shadow root seems to render anything https://codesandbox.io/p/sandbox/vanilla-js-forked-6964cz?file=%2Fsrc%2Findex.js%3A20%2C1

src/Shadowdomize/ShadowTreeWalker.ts
```diff
@@ -0,0 +1,272 @@
...
+
+                return this.previousNode();
+            } else {
+                return null;
+            }
+        }
+    }
+
+    public nextSibling(): Node | null {
+        throw new Error("Method not implemented.");
```
ling1726, line 253: we should only throw when not in production

ling1726 (approved):

src/Focusable.ts
mshoho: During the test I've discovered that native tree walker doesn't throw error if you set currentNode outside of the container. And my implementation had this constraint. So, I've just added the check as it seems to make sense.

src/Shadowdomize/DOMFunctions.ts
mshoho: Added TODO to index.ts.

src/Utils.ts
mshoho: This is a special case. If we need to insert dummy inputs around top-level element of a shadow root, they need to be inserted inside the shadow, not inside the host. Adding anything else to the shadow host (other than attached shadow) technically works, the elements appear in the DOM, but those elements are ignored by the browser, they are not rendered and they don't receive focus events.

src/Groupper.ts
mshoho: During the testing I've found that we might have issues when the groupper is created when the focus is somewhere inside. So, I've added it here to make sure the current groupper is set accordingly.

src/Shadowdomize/DOMFunctions.ts
mshoho: Yes. I have added it before the realization that browser ignores all children of the shadow host, other than attached shadow. We can remove this once these functions are moved to the dedicated package and covered with dedicated tests.

PR #351, opened 2024-01-24
Move focus APIs and custom events to move focus.

mshoho:
Adding `tabster:groupper:movefocus` and `tabster:mover:movefocus` for the application to move focus by triggering them.

### DIFF
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -39,7 +39,7 @@ export interface TabsterMoveFocusEventDetails {
     by: "mover" | "groupper" | "modalizer" | "root";
     owner: HTMLElement; // Mover, Groupper, Modalizer or Root, the initiator.
     next: HTMLElement | null; // Next element to focus or null if Tabster wants to go outside of Root (i.e. to the address bar of the browser).
-    relatedEvent: KeyboardEvent; // The original keyboard event that triggered the move.
+    relatedEvent?: KeyboardEvent; // The original keyboard event that triggered the move.
 }
 
 export type TabsterMoveFocusEvent =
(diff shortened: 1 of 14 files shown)


src/Types.ts
```diff
@@ -31,15 +38,21 @@ export const FocusableSelector = [
...
 
-export interface TabsterEventWithDetails<D> extends Event {
-    details: D;
-}
+// Trigger move focus event on a Mover element.
+export type MoverMoveFocusEvent = CustomEvent<{ key: MoverKey } | undefined>;
+
+// Enter or escape Groupper. Enter when `enter` is true, escape when `enter` is false.
+export type GroupperMoveFocusEvent = CustomEvent<
+    { enter: boolean } | undefined
```
ling1726, line 46: let's use an enum instead of a boolean so that the handling is clearer

src/Utils.ts
```diff
@@ -1794,6 +1797,20 @@ export function triggerMoveFocusEvent(
     return triggerEvent(details.owner, Types.MoveFocusEventName, details);
 }
 
+export function triggerMoverMoveFocusEvent(
```
ling1726, line 1800:
```suggestion
export function dispatchMoverMoveFocusEvent(
```

src/Types.ts
```diff
@@ -31,15 +38,21 @@ export const FocusableSelector = [
     "*[contenteditable]",
 ].join(", ");
 
-export interface TabsterEventWithDetails<D> extends Event {
-    details: D;
-}
+// Trigger move focus event on a Mover element.
+export type MoverMoveFocusEvent = CustomEvent<{ key: MoverKey } | undefined>;
```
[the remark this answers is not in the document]
mshoho: Mostly because it could be triggered by the application. So that the typings force to consider incorrect input.

src/Utils.ts
```diff
@@ -1794,6 +1797,20 @@ export function triggerMoveFocusEvent(
...
+    target: HTMLElement,
+    key: Types.MoverKey
+) {
+    return triggerEvent(target, Types.MoverMoveFocusEventName, { key });
+}
+
+export function triggerGroupperMoveFocusEvent(
```
ling1726, line 1807:
```suggestion
export function dispatchGroupperMoveFocusEvent(
```

src/Types.ts
```diff
@@ -798,7 +835,10 @@ export interface GroupperAPIInternal {
 }
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface GroupperAPI extends GroupperAPIInternal, Disposable {}
+export interface GroupperAPI extends GroupperAPIInternal, Disposable {
+    enterGroupper(element: HTMLElement): HTMLElement | null;
+    escapeGroupper(element: HTMLElement): HTMLElement | null;
```
ling1726, line 840: is there any reason for these to be public API since there are events that do these actions?

mshoho: I can actually have one moveFocus(action) method instead of two. It will be more future proof.

ling1726 (approved):

Issue #348, opened 2024-01-10
Support resetting memorizeCurrent for instances like new chat messages


micahgodbolt:
Right now, we memorize current in copilot chat so that if you tab down into input, but want to tab back up into chat, you don't lose your place.

The problem is that if a new message has been added to chance since you were last there, teams typically want focus to move to that most recent chat item.

defaultFocus only works if there is no memorized focus history, so that only works on the first focus.

https://github.com/microsoft/fluentai/issues/1307

PR #353, opened 2024-01-25
feat(mover): make isDefault always win against other props

smhigley:
Chatted with @ling1726 about this some time ago, this PR updates Mover to have `isDefault` always "win" when determining which element to focus when tabbing into a group.

This is related to work on Copilot messages (https://github.com/microsoft/tabster/issues/348), where we want `memorizeCurrent`, but sometimes need to override it, e.g. when a new message comes in.

### DIFF
--- a/src/Mover.ts
+++ b/src/Mover.ts
@@ -308,19 +308,19 @@ export class Mover
         ) {
             let found: HTMLElement | undefined | null;
 
-            if (memorizeCurrent) {
+            if (hasDefault) {
+                found = this._tabster.focusable.findDefault({
+                    container: moverElement,
+                    useActiveModalizer: true,
+                });
+            }
+
+            if (!found && memorizeCurrent) {
                 const current = this._current?.get();
 
                 if (current && state.acceptCondition(current)) {
                     found = current;
                 }
-            }
-
-            if (!found && hasDefault) {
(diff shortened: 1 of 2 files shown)


micahgodbolt:
if a user has 'isDefault' set, what is the best approach for actually using memorizeCurrent? Do you have to wait for focus, and then remove the 'isDefault' prop from the item? Do we need some callback to make sure we remove isDefault after focus has successfully been moved to the item?

smhigley:
@micahgodbolt yeah, I think the approach from the copilot side of things would be to remove `isDefault` once the element gets focus

mshoho:
This unconditionally alters the existing behaviour and might break some existing expectations. This behaviour either needs additional prop (like priority for memorizeCurrent) or the goal should be achieved differently (we were discussing custom DOM event like `tabster:mover-memorized` that you can dispatch to make Mover forget or alter memorized element (passing the desired action in the event details). @micahgodbolt what would be more convenient for your scenario? I think custom event would work better. I can do the adjustment today, this waits for too long already.

smhigley:
@mshoho the event approach sounds good to me, that sounds like it’ll address our use case.

If you let me know when the change is available, I’ll try it out in Fluent AI 👍

PR #362, opened 2024-02-27
Fixing races between parts of Tabster that want to focus something asynchronously.

mshoho:
Tabster does some focus movements asynchronously. Restorer restores lost focus asynchronously (to give the DOM the opportunity to finish the changes), Groupper handles Esc asynchronously (to give the application ability to handle Esc inside the Groupper).

This PR fixes the conflicts when both Groupper and Restorer want to move focus at the same time.

### DIFF
--- a/src/Groupper.ts
+++ b/src/Groupper.ts
@@ -430,6 +430,7 @@ export class GroupperAPI implements Types.GroupperAPI {
     private _current: Record<string, Types.Groupper> = {};
     private _grouppers: Record<string, Types.Groupper> = {};
     private _handleKeyPressTimer: number | undefined;
+    private _asyncFocusIntent: Types.AsyncFocusIntent | undefined;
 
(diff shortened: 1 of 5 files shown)


src/Groupper.ts
```diff
@@ -430,6 +430,7 @@ export class GroupperAPI implements Types.GroupperAPI {
     private _current: Record<string, Types.Groupper> = {};
     private _grouppers: Record<string, Types.Groupper> = {};
     private _handleKeyPressTimer: number | undefined;
+    private _asyncFocusIntent: Types.AsyncFocusIntent | undefined;
```
ling1726, line 433: Each entity only has **one** intent - there is no real reason why they need to use it to cancel and commit.

You could quite simply have `registerAsyncFocus(sourece, callback)` and `cancelAsyncFocus(source)` and have the FocusedElement state be the source of truth

PR #367, opened 2024-03-08
fix: hidden fixed elements do not receive focus

smhigley:
Related to #329

There's a bug where fixed elements that are hidden with `display: none`, or nested within another element with `display: none` were still treated as focusable.

This adds an additional check to `isDisplayNone` to catch fixed & hidden elements, as well as tests.

### DIFF
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -1915,7 +1915,10 @@ export function isDisplayNone(element: HTMLElement): boolean {
             return true;
         }
 
-        if (element.parentElement?.offsetParent === null && elementDocument.body !== element.parentElement) {
+        if (
+            element.parentElement?.offsetParent === null &&
+            elementDocument.body !== element.parentElement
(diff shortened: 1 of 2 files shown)


ling1726 (approved):

src/Utils.ts
```diff
@@ -1908,5 +1908,20 @@ export function isDisplayNone(element: HTMLElement): boolean {
...
     }
 
+    // if an element has display: fixed, we need to check if it is also hidden with CSS,
+    // or within a parent hidden with CSS
+    if (computedStyle?.position === "fixed") {
+        if (computedStyle.display === "none") {
+            return true;
+        }
+
+        if (
```
mshoho, line 1918: Instead of this `if` section, this should have used isDisplayNone on the parent element recursively. Without it the fix is incomplete for directly nested `position:fixed` elements.

PR #379, opened 2024-04-02
Improving how enums are declared.

mshoho:
We don't use native TypeScript enums because of how they are being compiled (into a non-treeshakeable result).
This PR adjusts how the enums are declared, making it more compact and reexportable while not changing it from the usage point of view.

### DIFF
--- a/src/CrossOrigin.ts
+++ b/src/CrossOrigin.ts
@@ -401,7 +401,7 @@ interface CrossOriginElementDataIn {
     /**
      * Optionally wait if the element is accessible or focusable before returning it
      */
-    accessibility?: Types.ObservedElementAccessibility;
+    accessibility?: Types.ObservedElementAccesibility;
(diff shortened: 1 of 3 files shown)


ling1726 (approved):

src/CrossOrigin.ts
```diff
@@ -401,7 +401,7 @@ interface CrossOriginElementDataIn {
     /**
      * Optionally wait if the element is accessible or focusable before returning it
      */
-    accessibility?: Types.ObservedElementAccesibility;
+    accessibility?: Types.ObservedElementAccessibility;
```
ling1726, line 404: FYI - the rename is a breaking change, please keep that in mind when bumping the next release

mshoho: Will do that adjustment as a separate PR.

PR #390, opened 2024-05-21
Rearranging exports so that types are separate from consts and consts are not bulk-exported as subproperty.

mshoho:
Better tree-shakeability FTW.


src/Consts.ts
```diff
@@ -0,0 +1,105 @@
...
+    Restorer: 2,
+    Deloser: 3,
+} as const;
+
+export const ObservedElementAccesibilities: Types.ObservedElementAccesibilities =
+    {
+        Any: 0,
+        Accessible: 1,
+        Focusable: 2,
+    } as const;
```
ling1726, line 35:
```suggestion
export const ObservedElementAccesibilities: Types.ObservedElementAccesibilities =
    {
        Any: 0,
        Accessible: 1,
        Focusable: 2,
    };
```

you've already typed this so there's no need to export as const

mshoho: Not really: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions — check the enum-like patterns section about const assertions.

mshoho: image
image

ling1726: but in the typings you can do equivalent with `readonly` 

https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAZiEMC8MDeAoG2YEMBcMAjADRY4BGhATGTjMIQMxkC+eED40GGoksCrgBOhAGIB5CSnTlsBYnUo0l2RjBYZWvAJZgoAU2FxcwAzEnTM9YQdwATcABsAnnkJEA3HJi2HztyoYam8bO0cwVwZmb20MIWEAOlwUZFRqDAQQZNT03gSc1CZvLMKNIA

ling1726: alternatively instead of declaring this twice you can do

import {  ObservedElementAccessibilities as  ObservedElementAccessibilitiesValues }  from './Const'

export type ObservedElementAccessibilities = typeof ObservedElementAccessibilitiesValues

src/EventsTypes.ts
```diff
@@ -0,0 +1,32 @@
...
+    relatedEvent?: KeyboardEvent; // The original keyboard event that triggered the move.
+}
+
+export type MoverMoveFocusEventDetail = { key: Types.MoverKey };
+
+export interface MoverMemorizedElementEventDetail {
+    memorizedElement: HTMLElement | undefined;
+}
+
+export type GroupperMoveFocusEventDetail = {
```
ling1726, line 21: why is there a mix of `type` and `interface`?

src/Focusable.ts
```diff
@@ -6,13 +6,14 @@
 import { getTabsterOnElement } from "./Instance";
 import { RootAPI } from "./Root";
 import * as Types from "./Types";
+import { FocusableSelector } from "./Consts";
```
ling1726, line 9: how about we adopt ALL_CAPS_KEBAB  as a convention for constants? that avoids a naming clash with the corresponding type and makes it clear in the code that constants are being used

src/Types.ts
```diff
@@ -145,14 +131,13 @@ export interface FocusedElementDetail {
...
-} as const;
+export interface AsyncFocusSources {
+    EscapeGroupper: 1;
+    Restorer: 2;
+    Deloser: 3;
+}
 
-export type AsyncFocusSource =
-    typeof AsyncFocusSources[keyof typeof AsyncFocusSources];
+export type AsyncFocusSource = AsyncFocusSources[keyof AsyncFocusSources];
```
ling1726, line 140: why is this type necessary, I have seen it used nowhere

ling1726: can this be removed without breaking the API?

mshoho: It is used just a few lines below in `requestAsyncFocus()` and `cancelAsyncFocus()` methods. Basically all of these singular types are for the enum-like pattern completeness. Singular type is for the variables and props of the enum type, plural hold possible values. `function goodFunc(prop: SingularType)` and `goodFunc(PluralType.Option1)`. With native enums it is the same name for both typing the props and picking values, but we have opted to not use native enums because of the nasty way they are transpiled.

src/Consts.ts
mshoho: Makes sense, adjusted.

src/Focusable.ts
mshoho: I've ALL_CAPS_KEBABed them.

src/EventsTypes.ts
mshoho: This is likely a leftover of event typings evolution. Adjusted to interfaces.

ling1726 (approved):

PR #391, opened 2024-05-28
getDummyInputContainer() utility function.

mshoho:
Sometimes it might be useful to be able to distinguish if an element is a dummy input in the application code. Exposing a function for that.

### DIFF
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -1925,3 +1925,18 @@ export function isDisplayNone(element: HTMLElement): boolean {
 
     return false;
 }
+
+/**
+ * If the passed element is Tabster dummy input, returns the container element this dummy input belongs to.
+ * @param element Element to check for being dummy input.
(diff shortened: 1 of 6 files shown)


src/Utils.ts
```diff
@@ -1925,3 +1925,18 @@ export function isDisplayNone(element: HTMLElement): boolean {
...
+ * If the passed element is Tabster dummy input, returns the container element this dummy input belongs to.
+ * @param element Element to check for being dummy input.
+ * @returns Dummy input container element (if the passed element is a dummy input) or null.
+ */
+export function getDummyInputContainer(
+    element: HTMLElement
+): HTMLElement | null {
+    return (
+        (
+            element as HTMLElementWithDummyContainer
```
bsunderhus, line 1939: nit: would is be better some safe guard method instead of casting?

bsunderhus (approved):

PR #395, opened 2024-07-25
feat(Mover): Connected Movers.

mshoho:
The ability to connect nested Movers via DOM hierarchy so that the parent Movers can continue handling the key press when the child Mover couldn't find new focus target.

### DIFF
--- a/tests/Mover.test.tsx
+++ b/tests/Mover.test.tsx
@@ -7,6 +7,7 @@ import * as React from "react";
 import {
     getTabsterAttribute,
     MoverDirections,
+    MoverConnections,
     MoverKeys,
     Types,
     Visibilities,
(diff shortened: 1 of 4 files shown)


tests/Mover.test.tsx
```diff
@@ -3145,3 +3146,366 @@ describe("Mover with virtual children provided by getParent()", () => {
...
+                            },
+                        })}
+                    >
+                        <div
+                            tabIndex={0}
+                            {...getTabsterAttribute({
+                                groupper: {},
+                                mover: {
+                                    direction: MoverDirections.Vertical,
+                                    connected: MoverConnections.Parent,
```
ling1726, line 3175: why are `Child` and `Parent` relationships both needed for connected movers? Is there a scenario where you can have a parent mover with `Child` and only a subset of movers will use `Parent`?

tests/Mover.test.tsx
```diff
@@ -3145,3 +3146,366 @@ describe("Mover with virtual children provided by getParent()", () => {
...
+                                    cyclic: true,
+                                },
+                            })}
+                        >
+                            <button>Button1</button>
+                            <button>Button2</button>
+                        </div>
+
+                        <div
+                            tabIndex={0}
```
ling1726, line 3185: what happens when a connected mover is not focusable?

tests/Mover.test.tsx
```diff
@@ -3145,3 +3146,366 @@ describe("Mover with virtual children provided by getParent()", () => {
...
+                                tabIndex={0}
+                                {...getTabsterAttribute({
+                                    groupper: {},
+                                    mover: {
+                                        direction: MoverDirections.Horizontal,
+                                        connected: MoverConnections.Parent,
+                                    },
+                                })}
+                            >
+                                <button>Button1</button>
```
ling1726, line 3292: Please update button names to be a bit more specific to their level of nesting

i.e. Button1L3

tests/Mover.test.tsx
```diff
@@ -3145,3 +3146,366 @@ describe("Mover with virtual children provided by getParent()", () => {
...
+    });
+
+    it("should move between more levels of connected Movers", async () => {
+        await new BroTest.BroTest(
+            (
+                <div {...getTabsterAttribute({ root: {} })}>
+                    <div
+                        {...getTabsterAttribute({
+                            mover: {
+                                direction: MoverDirections.Horizontal,
```
ling1726, line 3267: what happens when parent and child movers share the same direction?

ling1726 (approved):

PR #403, opened 2024-08-20
fix(modalizer): stop focus from leaving trapped modal w/ no focusable descendants

emmayjiang:
For a trapped modal with no focusable descendants, tabster should prevent focus from leaving it.

### DIFF
--- a/src/Modalizer.ts
+++ b/src/Modalizer.ts
@@ -272,6 +272,10 @@ export class Modalizer
                     findPropsOut
                 );
 
+                if (next === null) {
+                    next = currentElement;
+                }
+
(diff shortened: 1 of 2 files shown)


src/Modalizer.ts
```diff
@@ -272,6 +272,10 @@ export class Modalizer
                     findPropsOut
                 );
 
+                if (next === null) {
```
ling1726, line 275: Would be great to add a test for this, you can add it here and copy code from other tests to get started https://github.com/microsoft/tabster/blob/master/tests/Modalizer.test.tsx

You can run tests with `npm run test`

ling1726 (approved):

Issue #414, opened 2024-09-05
Expose internal scrollIntoView as public api?


acorncom:
I've been using Tabster on a client project over the last few months to noticeably improve their keyboard/accessibility story. Appreciate all the work you've put in on this library.

We hit a wrinkle when using it on a fairly complex table component (it's steadily growing toward Excel-like functionality). One of our use cases requires a custom event that prevents the `TabsterMoveFocusEvent` when doing page up / page down and uses our own logic instead. In the process, we lost the default `scrollIntoView` use [called by Mover](https://github.com/microsoft/tabster/blob/b5c269e6a355fb6c8b36aabc5759d6f4e773b08b/src/Mover.ts#L1209). Would you consider exposing the `scrollIntoView` util as public api?

### Issues seen

Amongst other things, the table component we're working with has a few custom features that snag simple mover page up / page down functionality:
- it allows scrolling left/right to see all columns
- it has a sticky "select" checkbox to the left of all columns

Here's a video showing a quick reproduction I built in the Tabster Storybook system:
https://github.com/user-attachments/assets/85c47ac5-0e28-4187-8217-1fa49c0a7036

In this case, the addition of the sticky "select" checkbox ends up causing bugs both with Move page up / page down and also Home / End when some columns are hidden off screen. That approach doesn't interact well with some of the assumptions built-in to the Mover page up/page down handler here [around relative positioning between elements](https://github.com/microsoft/tabster/blob/b5c269e6a355fb6c8b36aabc5759d6f4e773b08b/src/Mover.ts#L1046).

You can see that behavior on the reproduction PR I've just put up at https://github.com/microsoft/tabster/pull/415, see the "Mover -> Test Table with Mover Grid" example.

Would you want a patch for that behavior? Not sure if it'd be wise to change those assumptions for this particular use-case but if you have input on what you envision a patch looking like, we'd happily build that out.

### Our current approach

[...]

PR #415, opened 2024-09-05
Page up unexpected focus in a custom table - reproduction and proposed fix

acorncom:
This PR provides a reproduction of the issue written up in https://github.com/microsoft/tabster/issues/414.


acorncom:
@microsoft-github-policy-service agree

acorncom:
Per https://github.com/microsoft/tabster/issues/414#issuecomment-2361674198, we need to handle more than this, will close

Issue #413, opened 2024-09-05
[Feature request] Restorer source and target should be optionally linked with an identifier


ling1726:
Link restorer source and target elements with an optional identifier so that focus restore only happens between two types with the same identifier

Some pseudocode to illustrate desired behaviour

```
-> lose focus on source 'foo'
-> look through history
   -> target 'bar' <- ignore this
   -> target 'bar' <- ignore this
   -> target 'unknown' <-ignore this
   -> target 'foo' <- restore focus
```

```
-> lose focus on source 'unknown'
-> look through history
   -> target 'bar' <- ignore this
   -> target 'bar' <- ignore this
   -> target 'unknown' <- restore focus
   -> target 'foo' <- ignore this
```

PR #416, opened 2024-09-05
feature(restorer): link source and target by id

bsunderhus:
Introduces `id` attribute to `restorer`.

Providing an `id` will link a `source` with a `target` ensuring that focus restoration will only happen if the `id` is compatible between the `source` and the `target` element

Fixes https://github.com/microsoft/tabster/issues/413

### DIFF
--- a/src/Restorer.ts
+++ b/src/Restorer.ts
@@ -21,8 +21,6 @@ import {
 import { TabsterPart, WeakHTMLElement } from "./Utils";
 import { dom } from "./DOMAPI";
 
-const HISOTRY_DEPTH = 10;
-
 class Restorer extends TabsterPart<RestorerProps> implements RestorerInterface {
     private _hasFocus = false;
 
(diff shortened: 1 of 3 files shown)


src/Restorer.ts
```diff
@@ -77,9 +75,65 @@ class Restorer extends TabsterPart<RestorerProps> implements RestorerInterface {
...
+        const doc = this._getWindow().document;
+        for (let index = this._stack.length - 1; index >= 0; index--) {
+            const maybeElement = this._stack[index].get();
+            if (
+                !maybeElement ||
+                !dom.nodeContains(doc.body, dom.getParentElement(maybeElement))
+            ) {
+                this._stack.pop();
+            } else if (filter(maybeElement)) {
+                this._stack.splice(index, 1);
```
ling1726, line 126: I though you would have just continued popping, why do you think we should retain other history?

ling1726 (approved):

src/Restorer.ts
bsunderhus: Valid point, I was thinking it might be useful to not lose the restorer history while looking for a specific id. I guess we can just pop until we find the right one for now and then if this becomes necessary we can just introduce retention later on

PR #424, opened 2025-01-28
Fix(Modalizer): Adding activation history to the modalizer to activate the most recently active if the current one is removed.

mshoho:
If one modalizer is focused after another, then the second one is removed from DOM, the first one should be activated if it is still present.


src/Modalizer.ts
```diff
@@ -465,8 +467,38 @@ export class ModalizerAPI implements Types.ModalizerAPI {
...
+                const activationHistory = this._activationHistory;
+                const cleanActivationHistory: (string | undefined)[] = [];
+                let prevHistoryItem: string | undefined;
+
+                // The history order is from most recent to oldest.
+                for (let i = activationHistory.length; i--; ) {
+                    // Remove from activation history, making sure there are no duplicates
+                    // for cases like [modal2, modal1, modal2, modal1]: just removing modal2
+                    // will result in [modal1, modal1] and we want just [modal1].
+                    const mid = activationHistory[i];
```
ling1726, line 479: what is mid? can it be named more clearly?

src/Modalizer.ts
```diff
@@ -465,8 +467,38 @@ export class ModalizerAPI implements Types.ModalizerAPI {
...
 
+                const activationHistory = this._activationHistory;
+                const cleanActivationHistory: (string | undefined)[] = [];
+                let prevHistoryItem: string | undefined;
+
+                // The history order is from most recent to oldest.
+                for (let i = activationHistory.length; i--; ) {
+                    // Remove from activation history, making sure there are no duplicates
+                    // for cases like [modal2, modal1, modal2, modal1]: just removing modal2
+                    // will result in [modal1, modal1] and we want just [modal1].
```
ling1726, line 478: why is this deduplication important? each time you remove a modalizer, you remove every instance in the history and then select the most recent one to activate... so if the most recent one is duplicated won't it just be removed when itself is disposed?

mshoho: Basically, to not bloat this array over time. If you have a modalizer that sits there forever, and switch between it and other modalizers (that might come and go), the array will be becoming indefinitely bigger creating a memory lead. Yes, the case is narrow, the leak is marginal, but still, why not avoid it if it can be avoided. I can extend the comment with better explanation.

mshoho: Added a comment.

ling1726: I'm not really sure if this is a leak because you're just saving references to the same instance of a modalizer that should actually exist because it's still around.

It's your call

ling1726 (approved):

src/Modalizer.ts
mshoho: Activation history is not a modalizer instance property, it's a modalizer API property (so, it's a singleton) there is one history to rule them all.

mshoho: If you have a modal that is always in DOM, you activate it, it becomes a part of the activation history: [modal1]. Then you show another modal, activate it, it is added to the activation history: [modal2, modal1]. Then you don't close the second one, but just activate the first one: [modal1, modal2, modal1]. Then you activate second one again: [modal2, modal1, modal2, modal1]. Then you remove the second one: [modal1, modal1]. Repeat the process indefinitely and the array grows.

PR #425, opened 2025-02-03
feat(ObservedElement): Adding observed element request status.

mshoho:
When requestFocus() request resolves with `true` or `false`, in case of `false` it was not possible to tell the difference between timed out request or canceled request (older requests get canceled when a new request is made or when a user has moved focus). Adding status to be able to tell one from another.

### DIFF
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -231,7 +231,7 @@ export type ObservedElementRequestStatus =
 export interface ObservedElementAsyncRequest<T> {
     result: Promise<T>;
     cancel(): void;
-    status: ObservedElementRequestStatus;
+    status?: ObservedElementRequestStatus; // Making status optional for the interface backwards compatibility.
 }
 
 interface ObservedElementAPIInternal {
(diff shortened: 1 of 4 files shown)


src/Types.ts
```diff
@@ -221,9 +221,17 @@ export type ObservedElementAccessibilities =
...
+export type ObservedElementRequestStatuses =
+    typeof _ObservedElementRequestStatuses;
+
+export type ObservedElementRequestStatus =
+    ObservedElementRequestStatuses[keyof ObservedElementRequestStatuses];
+
 export interface ObservedElementAsyncRequest<T> {
     result: Promise<T>;
     cancel(): void;
+    status: ObservedElementRequestStatus;
```
ling1726, line 234: can you make this optional to avoid a breaking change on the interface?

ling1726 (approved):

src/Types.ts
mshoho: Done.

PR #437, opened 2025-02-28
fix(Modalizer): Fixing incorrect restore focus behaviour.

mshoho:
When an inactive modalizer was clicked and that click handler has moved focus back to active modalizer or when there is no focused element anymore, the restore focus logic was incorrectly trying to restore focus.


src/Modalizer.ts
```diff
@@ -1011,6 +1011,21 @@ export class ModalizerAPI implements Types.ModalizerAPI {
             return;
         }
 
+        const focusedElement = this._tabster.focusedElement.getFocusedElement();
+        const focusedElementModalizer =
+            focusedElement &&
+            RootAPI.getTabsterContext(this._tabster, focusedElement)?.modalizer;
+
+        if (
+            !focusedElement ||
```
ling1726, line 1020: does 'no focused element' mean that focus has landed on body?

Why is that a valid case not to do focus restoration? I would have throught if someone clicked outside of the modalizer on white space then the body would be focused

mshoho: It would be a valid case for Restorer or Deloser, but the modalizer's restoration purpose is to move focus back to active modalizer. And if the focus is on body, there is nowhere to move from. The modalizer focus restoration was always kicking onFocus of something new. I just haven't handled the cases when whatever has caused onFocus could have moving focus logic (or even intentional blur).

Particular TMP example: there is a popup, which is shown, but not active (focus is in the input outside of the popup). You click the popup to pick an option, that obviously moves focus inside the popup, but then the handler of this popup click moves focus back to the input. But the restore focus function has wrong reference element and steals focus into the wrong place. The same with intentional blur — if somebody has blurred intentionally, the restore focus logic will unblur it. Yes, focus on body might be a wrong thing to do, but that's not for modalizer to decide.

mshoho: Basically, all three new tests are about it. The third one tests intentional blur.

mshoho: I can even come up with a kinda legit reason for intentional blur — some custom deferred focus movement logic, similar to requestFocus(), allowing modalizer to refocus will give a conflict with the application logic.

PR #451, opened 2025-04-22
Avoid keyboard trap with disabled controls that have tabindex set

### LINKED ISSUE #399: Tab navigation not working on disabled elements with tabIndex=0
### Background:

There is [an issue in the Fluent UI repository](https://github.com/microsoft/fluentui/issues/30584) concerning tab navigation beyond disabled elements and [a PR with attempt to fix it](https://github.com/microsoft/fluentui/pull/32278). And it appears to be related to Tabster's handling of tab navigation for disabled elements with `tabindex="0"`.

### Steps to reproduce:
1. Visit https://codesandbox.io/p/sandbox/epic-easley-v2n9kf?file=%2Fsrc%2FApp.tsx%3A16%2C42
2. Use the tab key to navigate between buttons

### Expected result:
Tab navigation should bypass all disabled elements.

### Actual result:
Tab navigation halts after encountering disabled elements with `tabindex="0"`.

https://github.com/user-attachments/assets/0e7e5150-6b1d-4f4c-a0d1-d9af213961a9

**Note:** While removing `tabindex="0"` from disabled elements is a workaround, the question remains whether Tabster should inherently support such cases.

rgallagherab:
This pull request resolves an issue where a form control or button that can be truly disabled causes a keyboard trap if it is disabled and has `tabindex="0"`. Disabled form controls and buttons should never be considered focusable.

The selector `*[tabindex]` is not specific enough and causes checks against `FOCUSABLE_SELECTOR` to evaluate to true even if the control is also disabled.

Resolves #399

### DIFF
--- a/tests/Tabster.test.tsx
+++ b/tests/Tabster.test.tsx
@@ -609,4 +609,101 @@ describe("Tabster create", () => {
                 expect(el?.textContent).toEqual("Button5");
             });
     });
+
+    it("should skip disabled buttons and form controls with tabindex=0", async () => {
+        await new BroTest.BroTest(
+            (
+                <div
+                    id="items"
+                    {...getTabsterAttribute({
+                        root: {},
+                    })}
+                >
+                    <button>Button 1</button>
+                    <button disabled={true} tabIndex={0}>
+                        Button 2
+                    </button>
+                    <button>Button 3</button>
+                    <div
+                        {...getTabsterAttribute({
+                            mover: { cyclic: true, direction: 2 },
+                        })}
+                    >
+                        <button>Mover 1</button>
+                        <button disabled={true} tabIndex={0}>
+                            Mover 2
+                        </button>
+                        <button>Mover 3</button>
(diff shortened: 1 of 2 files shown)


tests/Tabster.test.tsx
```diff
@@ -609,4 +609,101 @@ describe("Tabster create", () => {
...
+                    {...getTabsterAttribute({
+                        root: {},
+                    })}
+                >
+                    <button>Button 1</button>
+                    <button disabled={true} tabIndex={0}>
+                        Button 2
+                    </button>
+                    <button>Button 3</button>
+                    <div
```
rgallagherab, line 627: This covers the case provided in the codesandbox example provided on the ticket, but that uses `controlTab: false`. When this is not the case any button or form component that has both `disabled` and `tabindex="0"` causes a keyboard trap.

rgallagherab:
@microsoft-github-policy-service agree company="AuditBoard"

PR #459, opened 2025-04-23
Massive deps bump

mshoho:


docs/package.json
```diff
@@ -39,7 +35,21 @@
         ]
     },
     "overrides": {
+        "bin-version": "^7.0.0",
```
layershifter, line 38: If everything is latest, why these are needed?

docs/package.json
```diff
@@ -11,20 +11,16 @@
...
     "dependencies": {
-        "@docusaurus/core": "2.0.0-beta.18",
-        "@docusaurus/preset-classic": "2.0.0-beta.18",
-        "@docusaurus/theme-live-codeblock": "2.0.0-beta.18",
-        "@mdx-js/react": "^1.6.21",
-        "clsx": "^1.1.1",
+        "@docusaurus/core": "3.7.0",
+        "@docusaurus/preset-classic": "3.7.0",
+        "@docusaurus/theme-live-codeblock": "3.7.0",
         "docusaurus": "^1.14.7",
```
layershifter, line 17: This should not be needed anymore

package.json
```diff
@@ -50,62 +50,72 @@
...
-        "tsconfig-paths-webpack-plugin": "^3.5.2",
-        "typescript": "^4.9.4",
-        "vite": "^2.9.12"
+        "tsconfig-paths-webpack-plugin": "^4.2.0",
+        "typescript": "^5.8.3",
+        "typescript-eslint": "^8.31.0",
+        "vite": "^6.3.2"
+    },
+    "optionalDependencies": {
+        "@rollup/rollup-linux-x64-gnu": "4.40.0"
```
layershifter, line 118: This should not be needed

package.json
```diff
@@ -50,62 +50,72 @@
...
     },
     "overrides": {
         "loader-utils": "^2.0.3"
```
layershifter, line 121: Can it be removed?

tests/utils/BroTest.ts
```diff
@@ -32,9 +31,46 @@ const domKey = process.env.SHADOWDOM ? "shadowDOM" : "dom";
...
+import { PassThrough } from "stream";
+import type {
+    PipeableStream,
+    RenderToPipeableStreamOptions,
+} from "react-dom/server";
+const renderToPipeableStream: (
+    element: React.ReactElement,
+    options?: RenderToPipeableStreamOptions
+) => PipeableStream =
+    require("../../node_modules/react-dom/cjs/react-dom-server.node.production.min").renderToPipeableStream;
```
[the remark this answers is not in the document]
layershifter: It's public API, https://react.dev/reference/react-dom/server/renderToPipeableStream

package.json
mshoho: This is needed because the build fails in the linux CI without it.

tests/utils/BroTest.ts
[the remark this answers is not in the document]
mshoho: I need minified version that doesn't have dev warnings like something is controlled/uncontrolled and needs some React props because of it. We use it for pure static HTML building and those warnings are irrelevant and create too much noise.

package.json
mshoho: You can see the error here: https://github.com/microsoft/tabster/actions/runs/14626618441/job/41039528037

tests/utils/BroTest.ts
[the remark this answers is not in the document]
layershifter: Is `onError` enough to silence them, https://react.dev/reference/react-dom/server/renderToPipeableStream#logging-crashes-on-the-server?

package.json
layershifter: Got it

docs/package.json
mshoho: Yay! It's now 0 in docs as well :).

.storybook/webpack.config.js
```diff
@@ -0,0 +1,30 @@
...
+                        "@babel/preset-react",
+                    ],
+                    plugins: [
+                        ["@babel/plugin-proposal-decorators", { legacy: true }],
+                        [
+                            "@babel/plugin-proposal-class-properties",
+                            { loose: true },
+                        ],
+                    ],
+                },
```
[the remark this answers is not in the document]
mshoho: I have picked that from Juraj's draft PR: https://github.com/microsoft/tabster/pull/450/files
But you are correct, deleted this file.

tests/utils/BroTest.ts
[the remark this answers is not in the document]
mshoho: Let me try.

azure-pipelines.yml
```diff
@@ -49,9 +49,9 @@ extends:
...
             
               - task: NodeTool@0
-                displayName: Use Node 16.x
+                displayName: Use Node 22.x
                 inputs:
-                  versionSpec: 16.x
+                  versionSpec: 22.x
               # PUPPETEER_DOWNLOAD_BASE_URL=https://storage.googleapis.com/chrome-for-testing-public is a workaround for puppeteer 
               # as puppeteer versions <22.0.0 will cause npm install to hang in CI environments.
               # see issue: https://github.com/puppeteer/puppeteer/issues/12833
```
layershifter, line 57: We can also remove this hack, right?

tests/utils/BroTest.ts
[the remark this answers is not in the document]
mshoho: Nah, I actually have this onError, the warning are still there:
      Warning: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`.
          at input
          at div
          at div

layershifter (approved):

PR #488, opened 2026-01-19
fix: handle inert in FocusableAPI

### LINKED ISSUE #369: Focusable does not handle the inert attribute
Tabster breaks on elements with `inert`, or nested within an ancestor with `inert`.

The easiest solution might be to add `:not([inert]):not([inert] *)` to the focusable selector, or add an explicit check (maybe in `isAccessible`?)

jonaphil:
- FOCUSABLE_SELECTOR now only matches if the element and all of its parents are not inert
- isFocusable returns false if element or its parents are inert
- isAccesible returns false if the element or its parents are inert

closes https://github.com/microsoft/tabster/issues/369


jonaphil:
@microsoft-github-policy-service agree company="networkteam GmbH"

PR #523, opened 2026-04-22
refactor: delegate attribute updates to each subsystem API

layershifter:
## Problem

`Instance.ts`'s `updateTabsterByAttribute()` held a ~1.3 kB minified switch that handled `data-tabster` attribute changes for every subsystem.

In a single-subsystem bundle the 5 dead branches for the unused subsystems shipped regardless, guarded only by runtime \`if (tabster.modalizer) …\` checks — which terser cannot statically prove unreachable.

## Fix

Move the heavy branch bodies (`create` / `setProps` / id-change recreation) onto each subsystem's API:

- `DeloserAPI.applyAttribute`
- `GroupperAPI.applyAttribute`
- `MoverAPI.applyAttribute`
- `ModalizerAPI.applyAttribute` — retains the id-change recreation path
- `RestorerAPI.applyAttribute`

\`Instance.ts\` branches become one-liners:

```ts
case "groupper":
    if (tabster.groupper) {
        tabster.groupper.applyAttribute(element, tabsterOnElement, props, sys);
    } else if (__DEV__) {
        console.error(\"Groupper API used before initialization, please call \\\`getGroupper()\\\`\");
    }
    break;
```

**Crucial effect:** each subsystem's `applyAttribute` body is now imported only when its module is imported — i.e. only when \`getX()\` is called. In a \`getGroupper\`-only bundle, Deloser / Mover / Modalizer / Restorer ship nothing for attribute handling; previously they shipped their create/setProps branches inside \`Instance.ts\`.

## Not changed

- **\`root\`** — always present, no tree-shaking benefit, kept inline.
- **\`observed\` / \`outline\`** — conditional but tiny (1-3 lines), kept inline.
- **\`focusable\` / \`uncontrolled\` / \`sys\`** — plain property assignments, kept inline.
- **The delete switch** (cleanup when a key disappears from the attribute) is already compact via case fall-through; a \`removeAttribute\` method would *add* bytes. Left alone.

### DIFF
--- a/src/Deloser.ts
+++ b/src/Deloser.ts
@@ -797,14 +797,14 @@ export class DeloserAPI implements Types.DeloserAPI {
 
     applyAttribute(
         element: HTMLElement,
-        storage: Types.TabsterOnElement,
+        existingDeloser: Types.Deloser | undefined,
         newProps: Types.DeloserProps
-    ): void {
-        if (storage.deloser) {
-            storage.deloser.setProps(newProps);
-        } else {
-            storage.deloser = this.createDeloser(element, newProps);
+    ): Types.Deloser {
+        if (existingDeloser) {
+            existingDeloser.setProps(newProps);
+            return existingDeloser;
         }
+        return this.createDeloser(element, newProps);
     }
 
     getActions(element: HTMLElement): Types.DeloserElementActions | undefined {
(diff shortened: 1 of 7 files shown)


mshoho (approved):

layershifter:
Alternative approach was merged.

PR #531, opened 2026-05-04
refactor(bundle-size): convert API classes to factory functions

layershifter:
## Summary

Convert API classes behind `get*` entry points (and `UncontrolledAPI`) from `class` to factory functions returning a closed-over public surface. Class scaffolding (constructor + private field declarations + arrow-bound method properties) doesn't mangle as well as plain top-level functions and closures, so each `getX` consumer ships less code.

8 APIs converted, one commit per API:

- `RestorerAPI`
- `OutlineAPI`
- `DeloserAPI`
- `GroupperAPI`
- `MoverAPI`
- `ModalizerAPI`
- `CrossOriginAPI`
- `UncontrolledAPI`

## Pattern

```ts
// before
export class RestorerAPI implements Types.RestorerAPI {
    private _tabster: TabsterCore;
    private _history: History;
    constructor(tabster: TabsterCore) { ... }
    dispose(): void { ... }
    private _onFocusIn = (...) => { ... };
    createRestorer(...) { ... }
}

// after
export function createRestorerAPI(tabster: TabsterCore): Types.RestorerAPI {
    const history = new History(tabster.getWindow);
    const onFocusIn = (...) => { ... };
    // ... wiring (event listeners, queueInit, subscriptions)
    return {
        createRestorer(...) { ... },
        dispose() { ... },
    };
}
```

Private fields → closure variables. Arrow-bound private methods → local `const` arrows. Constructor body → factory body. Public API → returned object. The interface in `Types.ts` is unchanged; consumers in `src/get/*.ts` swap `new XAPI(...)` for `createXAPI(...)`.

### Static methods

`DeloserAPI` had static methods (`getDeloser`, `getHistory`, `forceRestoreFocus`) used from `CrossOrigin.ts`. Kept under a same-named `const DeloserAPI = { ... }` namespace so call sites continue to work unchanged. Internal state needed by the statics is exposed via a `_`-prefixed internal interface.

## Bundle-size impact

Measured against `master` (post-#529). All `getX` fixtures shed bytes; per-API factory savings were ~570 B – 2.3 kB minified depending on API size.

| Fixture | Master | This PR | Δ min | Δ gzip |
|---|---|---|---|---|
| **all exports** | 113.84 kB / 30.94 kB | 105.63 kB / 30.17 kB | **−8.21 kB** | **−773 B** |
| createTabster (core) | 39.76 kB / 11.89 kB | 39.65 kB / 11.88 kB | −108 B | −13 B |
[...]

PR #538, opened 2026-05-11
chore(bundle-size): add fixtures for Focusable find* methods

layershifter:
## Summary

Adds four bundle-size fixtures used by the monosize harness — one per `Focusable` find variant, so each entry point can be measured independently:

- `focusableFindAll.fixture.js`
- `focusableFindLast.fixture.js`
- `focusableFindNext.fixture.js`
- `focusableFindPrev.fixture.js`

Purely additive — only files under `bundle-size/` are added, no shipped code changes.

The matching fixture for `getModalizer` with `controlTab: false` is deferred to the later stack PR that introduces `getRootDummyInputs`, since the fixture needs that export to exist.

## Stack context

First PR in the bundle-size stack. Parent: `master`.

### DIFF
--- a/bundle-size/focusableFindAll.fixture.js
+++ b/bundle-size/focusableFindAll.fixture.js
@@ -1,0 +1,23 @@
+import {
+    createTabster,
+    disposeTabster,
+    getTabsterAttribute,
+    setTabsterAttribute,
+    Types,
+} from "tabster";
+
+/** @param {ReturnType<typeof createTabster>} tabster */
+const useFocusable = (tabster) => tabster.focusable.findAll;
+
+console.log(
+    createTabster,
+    disposeTabster,
+    getTabsterAttribute,
+    setTabsterAttribute,
+    useFocusable,
+    Types
+);
+
+export default {
+    name: "focusable.findAll",
+};
--- a/bundle-size/focusableFindLast.fixture.js
+++ b/bundle-size/focusableFindLast.fixture.js
@@ -1,0 +1,23 @@
+import {
+    createTabster,
+    disposeTabster,
+    getTabsterAttribute,
+    setTabsterAttribute,
+    Types,
+} from "tabster";
+
+/** @param {ReturnType<typeof createTabster>} tabster */
+const useFocusable = (tabster) => tabster.focusable.findLast;
+
+console.log(
+    createTabster,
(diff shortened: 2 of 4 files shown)


layershifter:
## Stack overview

This PR is the first in a 14-PR stack splitting up the bundle-size refactor work (replaces #531). Land in order from top to bottom; each PR's diff against `master` is cumulative on top of every prior PR in the chain. Reviewers can subtract the parent's diff from the current PR's diff to see what's new in each one — each PR description has a "Stack context" section calling out the net new changes.

| # | PR | Title |
|---|---|---|
| 1 | this (#538) | chore(bundle-size): add fixtures for Focusable find* methods |
| 2 | #539 | chore(deps): patch keyborg to drop unused / legacy code paths |
| 3 | #540 | refactor(bundle-size): drop dead code paths + declare class fields |
| 4 | #541 | refactor(bundle-size): use free-function helpers for events and timers |
| 5 | #542 | refactor(bundle-size): convert Subscribable family to factories |
| 6 | #543 | refactor(bundle-size): convert DummyInput family to factories |
| 7 | #544 | refactor(bundle-size): convert API classes to factory functions |
| 8 | #545 | refactor(bundle-size): extract context resolvers + slim TabsterCore surface |
| 9 | #546 | refactor(bundle-size): extract Focusable mover/groupper resolver |
| 10 | #547 | refactor(bundle-size)!: hoist focusedElement helpers to module-level for tree-shaking |
| 11 | #548 | refactor(bundle-size): extract dummy infrastructure + auto-install via getRootDummyInputs |
| 12 | #549 | refactor(bundle-size)!: extract Tab keydown handler into opt-in module |
| 13 | #550 | refactor(bundle-size)!: drop UncontrolledAPI wrapper, store callback directly |
| 14 | #551 | refactor!: drop @deprecated API surface (keeps noDirectAriaHidden) |

After this PR lands on \`master\`, the remaining stack PRs need to be rebased onto \`master\` in turn. Doing so once #538 merges keeps each subsequent PR's diff focused on its own net change.

The final state at the tip of #551 matches the original \`refactor/api-factories\` branch (HEAD \`5f758a8\`) byte-for-byte, except for the bundle-size analyzer tooling under \`scripts/\` which is intentionally not shipped.

PR #560, opened 2026-06-11
refactor(bundle-size): use free-function helpers for DOM event APIs

layershifter:
## Summary

Introduces three single-character-mangleable wrappers in `Utils.ts` for the DOM event APIs whose property names would otherwise be preserved verbatim by the minifier:

- `addListener` / `removeListener` — thin wraps over `addEventListener` / `removeEventListener`, accept nullish targets.
- `dispatchEvent` — wraps `target.dispatchEvent(event)`, returns `false` for nullish targets (matching the prior `?.dispatchEvent` short-circuit shape that most call sites already handled).

All `src/**/*.ts` call sites converted (~71 sites across 11 files). Raw `.addEventListener` / `.removeEventListener` / `.dispatchEvent` property accesses now eliminated outside `Utils.ts`.

Enforced via a new `no-restricted-syntax` block in `eslint.config.mjs` scoped to `src/**/*.ts` (with `src/Utils.ts` exempted) so the wrappers can't drift back into raw calls in future PRs.

## Notes

A small structural tweak in `Deloser._scheduleRestoreFocus`: the prior `!curDeloserElement?.dispatchEvent(...)` short-circuit was rewritten as `if (!curDeloserElement) return; ...dispatchEvent(...)` to preserve type integrity now that the event constructor's `owner: HTMLElement` field is no longer protected by optional-chaining laziness. Same control flow as before.

## Stack context

A narrower companion to #541 — that PR also has these event-helper changes alongside the timer-helper conversions. Whichever lands first will leave the other a smaller diff after rebase.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -110,4 +110,33 @@ export default [
             ],
         },
     },
+    {
+        // Library source must go through the free-function helpers in
+        // `Utils.ts` instead of calling the raw DOM event APIs directly —
+        // helper names mangle to single chars after minification, native
+        // property names like `addEventListener` are preserved verbatim.
+        files: ["src/**/*.ts"],
+        ignores: ["src/Utils.ts"],
+        rules: {
+            "no-restricted-syntax": [
+                "error",
+                {
+                    selector:
+                        "CallExpression[callee.property.name='addEventListener']",
+                    message:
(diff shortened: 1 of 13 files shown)


layershifter:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #568, opened 2026-07-24
feat: support native focus options

petdud:
## Summary

- add `TabsterFocusOptions`, including `focusVisible` for TypeScript DOM versions that do not declare it yet
- allow `FocusedElementState.focus()` to accept and forward native focus options
- pass the same options through `ObservedElement.requestFocus()`
- preserve the existing boolean `preventScroll` argument for backwards compatibility

## Motivation

`HTMLElement.focus()` supports options beyond `preventScroll`, including `focusVisible`. Tabster currently exposes only a positional `preventScroll` boolean, so consumers cannot preserve valid native focus behavior when routing focus through Tabster.

## Compatibility

Existing calls such as `focusedElement.focus(element, undefined, undefined, true)` continue to work and are normalized to `{ preventScroll: true }`. Object options are forwarded unchanged.

`focusVisible` is declared explicitly in `TabsterFocusOptions` until it is available in all supported TypeScript DOM library versions.

## Validation

- `npm run format:check`
- `npm run lint:check`
- `npm run type-check:lib`
- `npm run type-check:tests`
- `npm run build-bundle`
- focused browser tests in normal DOM and Shadow DOM modes

> The initial bundle-size check failed because the stored `master` comparison artifact had expired (HTTP 410), not because of a bundle regression.

### DIFF
--- a/src/State/FocusedElement.ts
+++ b/src/State/FocusedElement.ts
@@ -204,7 +204,7 @@ export class FocusedElementState
         element: HTMLElement,
         noFocusedProgrammaticallyFlag?: boolean,
         noAccessibleCheck?: boolean,
-        preventScroll?: boolean
+        focusOptions?: boolean | Types.TabsterFocusOptions
     ): boolean {
         if (
             !this._tabster.focusable.isFocusable(
@@ -217,7 +217,12 @@ export class FocusedElementState
             return false;
         }
 
-        element.focus({ preventScroll });
+        const options =
+            typeof focusOptions === "boolean"
+                ? { preventScroll: focusOptions }
+                : focusOptions;
+
+        element.focus(options);
 
         return true;
     }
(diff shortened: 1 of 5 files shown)


petdud:
@Marata The remaining `bundle-size` failure is infrastructure-related: `monosize` is trying to download master artifact `7588359481`, which has expired with HTTP 410. The last successful **Bundle size Base** run on `master` was [27408103194](https://github.com/microsoft/tabster/actions/runs/27408103194) on June 12, and the workflow retains artifacts for 30 days. The workflow only triggers on pushes to `master` and neither of my accounts has Actions write permission. Could you rerun/refresh the master baseline workflow (or trigger a new master run), then rerun the PR bundle-size check?

mshoho (approved):

PR #588, opened 2026-08-22
fix: add exports map so ESM consumers resolve to the ESM build

### LINKED ISSUE #532: In v 8.8.0 package entry-point change breaks ESM named imports
tabster@8.7.0 works for us, but 8.8.0 started failing in Vitest.

Error:

`SyntaxError: Named export 'disposeTabster' not found. The requested module 'tabster' is a CommonJS module`

The regression seems to come from the package entrypoint change between these versions:

_verion 8.7.0_
```json
{
  "main": "./dist/index.js",
  "module": "./dist/tabster.esm.js"
}
```

_version 8.8.0_
```json
{
  "type": "module",
  "main": "./dist/cjs/index.cjs",
  "module": "./dist/esm/index.js"
}
```

In my case, an ESM consumer imports named exports from tabster, but in the Vitest / Node path it resolves through main, which is now CommonJS in v8.8.0.

It would help if tabster exposed explicit `exports` for `import` vs `require`, so Node-based tooling can resolve the ESM build reliably.
Something like this:
```json
"exports": {
  ".": {
    "types": "./dist/esm/index.d.ts",
    "import": "./dist/esm/index.js",
    "require": "./dist/cjs/index.cjs"
  }
}
```

dmytrokirpa:
Fixes #532

## Problem

Since 8.8.0 the package is `"type": "module"` with `"main"` pointing at the CJS bundle and no `"exports"` map. Node-based tooling (Vitest, Vite SSR) resolves `import { disposeTabster } from "tabster"` through `"main"` to `dist/cjs/index.cjs`, then asks `cjs-module-lexer` for the named-export list.

SWC routes named exports through a dynamic `_export(exports, {...})` helper plus `_export_star()`, neither of which the lexer can statically analyze. It reports zero named exports, so every named import fails:

```
SyntaxError: Named export 'disposeTabster' not found. The requested module 'tabster' is a CommonJS module
```

As noted in #532, this now affects every `@fluentui/react-*` consumer testing with Vitest, not just direct importers — `@fluentui/react-tabster@9.26.17` dropped its `"node": "./lib-commonjs/index.js"` condition, so Fluent's ESM build reaches `tabster` via `import` rather than `require`.

## Fix

Add the conditional `exports` map proposed in the issue. The `import` condition resolves to the real ESM build, where named exports are native and no lexing is involved. `require` keeps resolving to the CJS bundle, which works today because `require()` reads the live `exports` object at runtime.

`types` stays at `./dist/esm/index.d.ts` — the only directory declarations ship in — and `./package.json` remains exported so tooling can read it.

No build-script or output changes are needed; this is purely a resolution fix.

## Verification

Against a local `npm run build-bundle`:

| Path | Result |
| --- | --- |
| `import()` → `dist/esm/index.js` | 72 named exports, `createTabster` is a function |
| `require()` → `dist/cjs/index.cjs` | 72 runtime keys, `createTabster` is a function |

`cjs-module-lexer` still reports 0 exports for the CJS bundle, but that path is no longer reachable from an `import` statement once the `"import"` condition is present.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,14 @@
     "main": "./dist/cjs/index.cjs",
     "module": "./dist/esm/index.js",
     "types": "./dist/esm/index.d.ts",
+    "exports": {
+        ".": {
+            "types": "./dist/esm/index.d.ts",
+            "import": "./dist/esm/index.js",
+            "require": "./dist/cjs/index.cjs"
+        },
+        "./package.json": "./package.json"
+    },
     "files": [
         "dist"
     ],
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
             "import": "./dist/esm/index.js",
             "require": "./dist/cjs/index.cjs"
         },
+        "./dist/*": "./dist/*",
         "./package.json": "./package.json"
     },
     "files": [


Hotell (approved):
lets add a automation as follow up to guarantee this works as expected going forward. ty

Hotell (approved):
LGTM

mshoho (approved):

